PR #16, opened 2024-06-19
NamespaceId, SubspaceId, PayloadDigest, IsAuthorisedWrite, Entry, AuthorisedEntry

sgwilym:
- Added `NamespaceId` trait
- Added `SubspaceId` trait
- Added `PayloadDigest` trait
- `Entry` is just a struct, rather than a trait.
  - Just uses normal generics rather than const generics, as these are constrained to a few types. 
- `AuthorisedEntry` is a struct too.
  - There is no `PossiblyAuthorisedEntry`.
  - `AuthorisedEntry::new` uses the `IsAuthorisedWrite` trait (implementing `Fn`) to determine whether it should return an error or not.


willow-data-model/src/data_model.rs
```diff
@@ -1 +1,2 @@
+pub mod entry;
```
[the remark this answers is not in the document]
sgwilym: I'm looking for a nice shorthand for this, after having unsuccessfully applied 'read more' to many docs. It would be nice to have something I can drop in at the end of a sentence or paragraph like "Specification definition". But _catchier_.

willow-data-model/src/data_model/entry.rs
```diff
@@ -0,0 +1,95 @@
...
+impl<NamespaceId, SubspaceId, P, PayloadDigest> Entry<NamespaceId, SubspaceId, P, PayloadDigest>
+where
+    NamespaceId: PartialEq + Eq,
+    SubspaceId: PartialOrd + Ord + PartialEq + Eq,
+    P: Path,
+    PayloadDigest: PartialOrd + Ord + PartialEq + Eq,
+{
+    /// Return if this [`Entry`] is newer than another using their timestamps.
+    /// Tie-breaks using the Entries' payload digest and payload length otherwise.
+    pub fn is_newer_than(&self, other: &Entry<NamespaceId, SubspaceId, P, PayloadDigest>) -> bool {
```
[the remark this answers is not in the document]
sgwilym: yeah its fine, it's no typescript though

willow-data-model/src/data_model.rs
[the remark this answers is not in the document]
AljoschaMeyer: "Reference, Definition, implements, Specification, Spec"?

willow-data-model/src/data_model/parameters.rs
```diff
@@ -0,0 +1,42 @@
+use super::{
+    entry::{AuthorisedEntry, Entry, UnauthorisedWriteError},
+    path::Path,
+};
+
+/// A type for identifying [namespaces](https://willowprotocol.org/specs/data-model/index.html#namespace).
+/// [Read more](https://willowprotocol.org/specs/data-model/index.html#NamespaceId).
+pub trait NamespaceId: PartialEq + Eq {
+    /// A [`NamespaceId`] for when we have to provide one but its meaning doesn't matter.
+    const DEFAULT_NAMESPACE_ID: Self;
```
sgwilym, line 10: I should totally implement the `Default` trait for this instead.

willow-data-model/src/data_model/entry.rs
```diff
@@ -0,0 +1,96 @@
...
+    parameters::{IsAuthorisedWrite, NamespaceId, PayloadDigest, SubspaceId},
+    path::Path,
+};
+
+/// A Timestamp is a 64-bit unsigned integer, that is, a natural number between zero (inclusive) and 2^64 - 1 (exclusive).
+/// Timestamps are to be interpreted as a time in microseconds since the Unix epoch.
+/// [Read more](https://willowprotocol.org/specs/data-model/index.html#Timestamp).
+pub type Timestamp = u64;
+
+#[derive(Debug, PartialEq, Eq, Clone)]
```
mycognosist, line 11: This `#[derive(Debug, PartialEq, Eq, Clone)]` should go below the doc comments (on the line above `pub struct Entry<N, S, P, PD>`. Same with all other cases.

willow-data-model/src/data_model/parameters.rs
AljoschaMeyer: Also, the doc comment should specify that this is not an arbitrary selection up to the implementor, but spec-mandated.

sgwilym: There are only constraints on `SubspaceId`, I think (that it must be the least element in the total order of `SubspaceId`). I can't find anything that suggests the default namespace ID or payload digest need to satisfy any particular requirements.

AljoschaMeyer: Huh, you are right. You keep learning new things if you also forget things quickly enough!

data-model/src/entry.rs
```diff
@@ -0,0 +1,182 @@
...
+    parameters::{IsAuthorisedWrite, NamespaceId, PayloadDigest, SubspaceId},
+    path::Path,
+};
+
+/// A Timestamp is a 64-bit unsigned integer, that is, a natural number between zero (inclusive) and 2^64 - 1 (exclusive).
+/// Timestamps are to be interpreted as a time in microseconds since the Unix epoch.
+/// [Read more](https://willowprotocol.org/specs/data-model/index.html#Timestamp).
+pub type Timestamp = u64;
+
+/// The metadata associated with each Payload.
```
AljoschaMeyer, line 11:
```suggestion
/// The metadata associated with each Payload.
/// [Read more](https://willowprotocol.org/specs/data-model/index.html#Entry).
```

data-model/src/entry.rs
```diff
@@ -0,0 +1,182 @@
...
+            || (other.timestamp == self.timestamp && other.payload_digest < self.payload_digest)
+            || (other.timestamp == self.timestamp
+                && other.payload_digest == self.payload_digest
+                && other.payload_length < self.payload_length)
+    }
+}
+
+/// An error indicating an [`AuthorisationToken`] does not authorise the writing of this entry.
+#[derive(Debug)]
+pub struct UnauthorisedWriteError;
```
AljoschaMeyer, line 62: Would expect this to implement the `Error` trait. Which... is a bit of an involved topic, unfortunately =S

data-model/src/entry.rs
```diff
@@ -0,0 +1,182 @@
...
+    PD: PayloadDigest;
+
+impl<N, S, P, PD, AT> AuthorisedEntry<N, S, P, PD, AT>
+where
+    N: NamespaceId,
+    S: SubspaceId,
+    P: Path,
+    PD: PayloadDigest,
+{
+    /// Construct an [`AuthorisedEntry`] if the token permits the writing of this entry, otherwise return a [`UnauthorisedWriteError`]
```
AljoschaMeyer, line 88:
```suggestion
    /// Construct an [`AuthorisedEntry`] if the token permits the writing of this entry, otherwise return an [`UnauthorisedWriteError`]
```

data-model/src/parameters.rs
```diff
@@ -0,0 +1,38 @@
...
+    entry::{AuthorisedEntry, Entry, UnauthorisedWriteError},
+    path::Path,
+};
+
+/// A type for identifying [namespaces](https://willowprotocol.org/specs/data-model/index.html#namespace).
+/// [Definition](https://willowprotocol.org/specs/data-model/index.html#NamespaceId).
+
+pub trait NamespaceId: Eq + Default {}
+
+/// A type SubspaceId for identifying [subspaces](https://willowprotocol.org/specs/data-model/index.html#subspace).
```
AljoschaMeyer, line 11: Either remove third word or add "NamespaceId" as third word of the doc comment for NamespaceId.

data-model/src/parameters.rs
```diff
@@ -0,0 +1,38 @@
...
+pub trait SubspaceId: Ord + Default {}
+
+/// A totally ordered type for [content-addressing](https://en.wikipedia.org/wiki/Content_addressing) the data that Willow stores.
+/// [Definition](https://willowprotocol.org/specs/data-model/index.html#PayloadDigest).
+pub trait PayloadDigest: Ord + Default {}
+
+/// A function that maps an [`Entry`] and an [`AuthorisationToken` (willowprotocol.org)](https://willowprotocol.org/specs/data-model/index.html#AuthorisationToken) to a result indicating whether the `AuthorisationToken` does prove write permission for the [`Entry`].
+/// [Definition](https://willowprotocol.org/specs/data-model/index.html#is_authorised_write).
+///
+/// This function intentionally deviates from the specification's definition (in which `is_authorised_write` returns a `bool`) so that the result of the function cannot be accidentally ignored.
```
AljoschaMeyer, line 26: Couple notes here. A function named `is_whatever` should return a `bool`, pretty much irrespective of the context. When I read `AuthorisedEntry::new` and got thoroughly confused for a second.

I think I'd prefer if this parameter was a simple predicate, in accordance with the spec. In `AuthorisedEntry::new`, you'd be calling that predicate and return a result according to that. As long as `AuthorisedEntry::new` is the only way of obtaining AuthorisedEntries, then you get both the type safety and closeness to the spec in the parameter defs.

data-model/src/entry.rs
sgwilym: Is this a blocker?

sgwilym: Going to merge this as we have a number of other errors which do not implement the `Error` trait, and we can do that all in one go.

PR #19, opened 2024-06-21
Implement RangeEnd, Range

sgwilym:
Implements Willow’s one-dimensional ‘range’ grouping [as described here](https://willowprotocol.org/specs/grouping-entries/index.html#range), with testing for value inclusion, range inclusion, intersection creation, and a method which creates a range including all values of that type.

While it's still possible to create empty ranges by initialising a `Range` directly, attempting to do so with `new_closed` will return a `EmptyRangeError`.

This is a draft while I add some ~~fuzzing~~ tests, but I think the implementation is ready to look at!


data-model/src/groupings/range.rs
```diff
@@ -0,0 +1,159 @@
+use std::cmp::Ordering;
+
+#[derive(Debug, PartialEq, Eq)]
+/// Determines whether a [`Range`] is _closed_ or _open_.
+pub enum RangeEnd<T>
+where
+    T: PartialEq + Eq + PartialOrd + Ord,
```
AljoschaMeyer, line 7: `Ord` implies the other three, so you can remove those (here and pretty much everywhere =))

data-model/src/groupings/range.rs
```diff
@@ -0,0 +1,159 @@
...
+
+/// One-dimensional grouping over a type of value.
+///
+/// [Definition](https://willowprotocol.org/specs/grouping-entries/index.html#range)
+#[derive(Debug, PartialEq, Eq)]
+pub struct Range<T>
+where
+    T: PartialEq + Eq + PartialOrd + Ord,
+{
+    /// A range [includes](https://willowprotocol.org/specs/grouping-entries/index.html#range_include) all values greater than or equal to its [start value](https://willowprotocol.org/specs/grouping-entries/index.html#start_value).
```
AljoschaMeyer, line 60: Technically not true, since it filters out most of those with the end value again. That's why the spec uses this slightly unwieldy "To be included in a range, a value must be gt the start value" construction at times (iirc).

data-model/src/groupings/range.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    /// Returns `true` if `other` range is fully [included](https://willowprotocol.org/specs/grouping-entries/index.html#range_include) in this [`Range`].
+    pub fn includes_range(&self, other: &Range<T>) -> bool {
+        self.start <= other.start && self.end >= other.end
+    }
+
+    /// Create the [intersection](https://willowprotocol.org/specs/grouping-entries/index.html#range_intersection) between this [`Range`] and another `Range`.
+    pub fn intersection(&self, other: &Self) -> Option<Self> {
+        let start = (&self.start).max(&other.start);
```
AljoschaMeyer, line 106:
```suggestion
        let start = cmp::max(&self.start, &other.start);
```

Is more idiomatic I think.

data-model/src/groupings/range.rs
```diff
@@ -0,0 +1,159 @@
+use std::cmp::Ordering;
```
AljoschaMeyer, line 1: In general, try to import from `core::*` instead of `std::*` for this project; that will simplify nostd things if/when we get to them.

AljoschaMeyer:
Btw, I think that all functions that deal with potentially empty ranges should either use Option<Range> or Result<Range, RangeError>, but no mixture (currently, intersecting yields an Option), but closed creation a Result.

sgwilym:
@AljoschaMeyer sure. I think it makes more sense for creation to be an `Option` than to make non-intersection return an error, on it.

PR #20, opened 2024-06-23
Range3d, Area, and AreaOfInterest

sgwilym:
## Implement `Range3d`

- Added `Range3d` struct with `includes_entry` and `intersection` methods, `Default` impl.
- `Range::includes` no longer moves the value being tested for inclusion
- Add Clone trait bound to NamespaceId, SubspaceId, PayloadDigest

## Implement `AreaSubspace`, `Area`

- Add `AreaSubspace` with `includes` and `intersection` methods
- Add `Area` with `full`, `subspace`, `includes_entry`, and `intersection` methods.

## Implement `AreaOfInterest`

- Add `AreaOfInterest` with `intersection` method.

### DIFF
--- a/data-model/src/grouping/area.rs
+++ b/data-model/src/grouping/area.rs
@@ -1,0 +1,273 @@
+use crate::{
+    entry::{Entry, Timestamp},
+    parameters::{NamespaceId, PayloadDigest, SubspaceId},
+    path::Path,
+};
+
+use super::{range::Range, range_3d::Range3d};
+
+/// The possible values of an [`Area`]'s `subspace`.
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub enum AreaSubspace<S: SubspaceId> {
+    /// A value that signals that an [`Area`] includes Entries with arbitrary subspace_ids.
+    Any,
+    /// A concrete [`SubspaceId`]
+    Id(S),
+}
+
+impl<S: SubspaceId> AreaSubspace<S> {
+    /// Whether this [`AreaSubspace`] includes a given [`SubspaceId`].
+    pub fn includes(&self, sub: &S) -> bool {
+        match self {
+            AreaSubspace::Any => true,
+            AreaSubspace::Id(id) => sub == id,
+        }
+    }
+
+    /// Return the intersection between two [`AreaSubspace`].
+    fn intersection(&self, other: &Self) -> Option<Self> {
+        match (self, other) {
+            (Self::Any, Self::Any) => Some(Self::Any),
+            (Self::Id(a), Self::Any) => Some(Self::Id(a.clone())),
+            (Self::Any, Self::Id(b)) => Some(Self::Id(b.clone())),
+            (Self::Id(a), Self::Id(b)) if a == b => Some(Self::Id(a.clone())),
+            (Self::Id(_a), Self::Id(_b)) => None,
+        }
+    }
+}
+
+/// A grouping of entries.
+/// [Definition](https://willowprotocol.org/specs/grouping-entries/index.html#areas).
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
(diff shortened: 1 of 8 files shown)


data-model/src/grouping/area.rs
```diff
@@ -0,0 +1,273 @@
...
+        self.subspace.includes(&entry.subspace_id)
+            && self.path.is_prefix_of(&entry.path)
+            && self.times.includes(&entry.timestamp)
+    }
+
+    /// Return the intersection of this [`Area`] with another.
+    /// [Definition](https://willowprotocol.org/specs/grouping-entries/index.html#area_intersection).
+    pub fn intersection(&self, other: &Area<S, P>) -> Option<Self> {
+        let subspace_id = self.subspace.intersection(&other.subspace)?;
+        let path = if self.path.is_prefix_of(&other.path) {
```
AljoschaMeyer, line 86: Isn't this equivalent to taking the greatest common prefix of the two paths? Does `Path` have a function for that?

data-model/src/grouping/area.rs
```diff
@@ -0,0 +1,273 @@
...
+        let times = self.times.intersection(&other.times)?;
+        Some(Self {
+            subspace: subspace_id,
+            times,
+            path,
+        })
+    }
+
+    /// Return a [`Range3d`] with equivalent inclusion to this [`Area`].
+    pub fn into_range(&self) -> Range3d<S, P> {
```
AljoschaMeyer, line 102: Could (should?) (also) implement `From<Area>` for `Range3d`.

data-model/src/grouping/area.rs
```diff
@@ -0,0 +1,273 @@
+use crate::{
+    entry::{Entry, Timestamp},
+    parameters::{NamespaceId, PayloadDigest, SubspaceId},
+    path::Path,
+};
+
+use super::{range::Range, range_3d::Range3d};
```
mycognosist, line 7: Suggest rather importing these as part of the `use crate::` section above. I generally only ever see `super` used in tests.

data-model/src/grouping/range_3d.rs
```diff
@@ -0,0 +1,255 @@
...
+            subspaces,
+        })
+    }
+}
+
+impl<S: SubspaceId, P: Path> Default for Range3d<S, P> {
+    /// [Definition](https://willowprotocol.org/specs/grouping-entries/index.html#default_3d_range).
+    fn default() -> Self {
+        Self {
+            subspaces: Default::default(),
```
AljoschaMeyer, line 51: Can you do something along the lines of `S::default()` here?

data-model/src/grouping/area_of_interest.rs
```diff
@@ -0,0 +1,29 @@
+use crate::{parameters::SubspaceId, path::Path};
+
+use super::area::Area;
```
mycognosist, line 3: Rather import these as part of the `use crate::` section above.

data-model/src/grouping/range_3d.rs
```diff
@@ -0,0 +1,255 @@
+use super::range::Range;
```
mycognosist, line 1: Same comment about `use crate::`.

data-model/src/parameters.rs
```diff
@@ -1,24 +1,21 @@
-use super::{
-    entry::{AuthorisedEntry, Entry, UnauthorisedWriteError},
-    path::Path,
-};
+use super::{entry::Entry, path::Path};
```
mycognosist, line 1: `use crate::`

data-model/src/grouping/area.rs
sgwilym: It's not _quite_ that - it's taking the more specific of the two paths if one is a prefix of the other.

AljoschaMeyer: Ah, right, I was confusing this with union. Again. 🫠

PR #21, opened 2024-06-24
Store trait

sgwilym:
Work in progress, currently being discussed and iterated upon.

### DIFF
--- a/data-model/src/store.rs
+++ b/data-model/src/store.rs
@@ -1,0 +1,104 @@
+use crate::{
+    entry::{AuthorisedEntry, Entry},
+    grouping::{area::Area, area_of_interest::AreaOfInterest, range_3d::Range3d},
+    parameters::{NamespaceId, PayloadDigest, SubspaceId},
+    path::Path,
+};
+
+// eheheheh
+type Payload = String;
+
+/// Returned when an entry is successfully ingested into a [`Store`].
+pub enum EntryIngestionSuccess<
+    N: NamespaceId,
+    S: SubspaceId,
+    P: Path,
+    PD: PayloadDigest,
+    AT,
+    AEI: Iterator<Item = AuthorisedEntry<N, S, P, PD, AT>>,
+> {
+    /// The entry was successfully ingested.
+    Success(AuthorisedEntry<N, S, P, PD, AT>),
+    /// The entry was successfully ingested and prefix pruned some entries.
+    SuccessAndPruned(AEI),
+}
+
+pub enum EntryIngestionError<N: NamespaceId, S: SubspaceId, P: Path, PD: PayloadDigest, AT> {
+    NewerPrefix {
+        /// The obsolete entry which is pruned by the newer entry.
+        pruned: AuthorisedEntry<N, S, P, PD, AT>,
+        /// The newer entry whose path prefixes the obsolete entry's.
+        newer: AuthorisedEntry<N, S, P, PD, AT>,
+    },
+    Obsolete {
+        /// The obsolete entry which was not ingested.
+        obsolete: AuthorisedEntry<N, S, P, PD, AT>,
+        /// The newer entry which was not overwritten.
+        newer: AuthorisedEntry<N, S, P, PD, AT>,
+    },
+    /// The entry belonged to another namespace.
+    WrongNamespace(AuthorisedEntry<N, S, P, PD, AT>),
+}
+
+/// Returned when a [`Payload`] fails to be ingested into the [`Store`the ]
+pub enum PayloadIngestionError {
+    /// None of the entries in the store reference this payload.
+    NotEntryReference,
+    /// The payload is already held in storage.
+    AlreadyHaveIt,
+    /// The received's payload digest is not what was expected.
+    DigestMismatch,
+    /// Try deleting some files!!!
+    SomethingElseWentWrong,
+}
+
+pub enum QueryOrder {
+    Subspace,
+    Path,
+    Timestamp,
+}
+
+/// A [`Store`] is a set of [`AuthorisedEntry`].
+pub trait Store<
+    N: NamespaceId,
+    S: SubspaceId,
+    P: Path,
+    PD: PayloadDigest,
+    AT,
+    AEI: Iterator<Item = AuthorisedEntry<N, S, P, PD, AT>>,
+    // Do we need a trait for fingerprints?
+    FP,
+>
+{
+    /// Attempt to store an [`AuthorisedEntry`] in the [`Store`].
+    fn ingest_entry(
+        &self,
+        expected_digest: PD,
+        authorised_entry: AuthorisedEntry<N, S, P, PD, AT>,
+    ) -> Result<EntryIngestionSuccess<N, S, P, PD, AT, AEI>, EntryIngestionError<N, S, P, PD, AT>>;
+    // I'm told this is too complex, but I'm also told trait bounds can't be enforced on type aliases.
+
+    /// Attempt to ingest a payload for a given [`AuthorisedEntry`].
+    fn ingest_payload(&self, payload: Payload) -> Result<(), PayloadIngestionError>;
+
+    /// Query the store's set of [`AuthorisedEntry`] using an [`Area`].
+    fn query_area(&self, area: Area<S, P>) -> AEI;
+
+    /// Query the store's set of [`AuthorisedEntry`] using a [`Range3d`].
+    fn query_range(&self, area: Range3d<S, P>, order: QueryOrder, reverse: bool) -> AEI;
+
+    /// Try to get the payload for a given [`Entry`].
+    fn getPayload(&self, entry: &Entry<N, S, P, PD>) -> Option<Payload>;
+
+    /// Summarise a given range into a fingerprint.
+    fn summarise(&self, range: Range3d<S, P>) -> FP;
+
+    /// Split a range into two approximately equally sized ranges, or return the original range if it can't be split.
+    fn split_range(
+        &self,
+        range: Range3d<S, P>,
+    ) -> Result<(Range3d<S, P>, Range3d<S, P>), Range3d<S, P>>;
+
+    /// Transform an [`AreaOfInterest`] into a concrete range by using its `max_count` and `max_size` properties as clamps.
+    fn aoi_to_range(&self, aoi: AreaOfInterest<S, P>) -> Range3d<S, P>;
+}
--- a/data-model/src/store.rs
+++ b/data-model/src/store.rs
@@ -1,3 +1,5 @@
+use std::future::Future;
+
 use ufotofu::nb::BulkProducer;
 
 use crate::{
@@ -91,10 +93,14 @@ where
     fn ingest_entry(
         &self,
         authorised_entry: AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>,
-    ) -> Result<
-        EntryIngestionSuccess<MCL, MCC, MPL, N, S, PD, AT>,
-        EntryIngestionError<MCL, MCC, MPL, N, S, PD, AT>,
+    ) -> impl Future<
+        Output = Result<
+            EntryIngestionSuccess<MCL, MCC, MPL, N, S, PD, AT>,
+            EntryIngestionError<MCL, MCC, MPL, N, S, PD, AT>,
+        >,
     >;
+
+    // TODO: Bulk ingestion entry. The annoying there is it needs its own success / error types. We could get around this by exposing a BulkConsumer from the Store, but [then we need to support multi-writer consumption](https://github.com/earthstar-project/willow-rs/pull/21#issuecomment-2192393204).
 
     /// Attempt to append part of a payload for a given [`AuthorisedEntry`].
     ///
@@ -111,7 +117,7 @@ where
         expected_digest: PD,
         expected_size: u64,
         producer: &mut Producer,
-    ) -> Result<PayloadAppendSuccess<MCL, MCC, MPL, N, S, PD>, PayloadAppendError>
+    ) -> impl Future<Output = Result<PayloadAppendSuccess<MCL, MCC, MPL, N, S, PD>, PayloadAppendError>>
     where
         Producer: BulkProducer<Item = u8>;
 }
--- a/data-model/src/store.rs
+++ b/data-model/src/store.rs
@@ -41,19 +41,31 @@ pub enum EntryIngestionError<
     S: SubspaceId,
     PD: PayloadDigest,
     AT,
+    OE,
 > {
     /// The entry belonged to another namespace.
     WrongNamespace(AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>),
     /// The ingestion would have triggered prefix pruning when that was not desired.
     PruningPrevented,
+    /// Something specific to this store implementation went wrong.
+    OperationsError(OE),
 }
 
 /// A tuple of an [`AuthorisedEntry`] and how a [`Store`] responded to its ingestion.
-pub type BulkIngestionResult<const MCL: usize, const MCC: usize, const MPL: usize, N, S, PD, AT> = (
+pub type BulkIngestionResult<
+    const MCL: usize,
+    const MCC: usize,
+    const MPL: usize,
+    N,
+    S,
+    PD,
+    AT,
+    OE,
+> = (
     AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>,
     Result<
         EntryIngestionSuccess<MCL, MCC, MPL, N, S, PD, AT>,
-        EntryIngestionError<MCL, MCC, MPL, N, S, PD, AT>,
+        EntryIngestionError<MCL, MCC, MPL, N, S, PD, AT, OE>,
     >,
 );
 
@@ -66,9 +78,10 @@ pub struct BulkIngestionError<
     S: SubspaceId,
     PD: PayloadDigest,
     AT: AuthorisationToken<MCL, MCC, MPL, N, S, PD>,
+    OE,
     IngestionError,
 > {
-    pub ingested: Vec<BulkIngestionResult<MCL, MCC, MPL, N, S, PD, AT>>,
+    pub ingested: Vec<BulkIngestionResult<MCL, MCC, MPL, N, S, PD, AT, OE>>,
     pub error: IngestionError,
 }
 
@@ -86,7 +99,7 @@ where
 }
 
 /// Returned when a payload fails to be appended into the [`Store`].
-pub enum PayloadAppendError {
+pub enum PayloadAppendError<OE> {
     /// None of the entries in the store reference this payload.
     NotEntryReference,
     /// The payload is already held in storage.
@@ -95,8 +108,8 @@ pub enum PayloadAppendError {
     PayloadTooLarge,
     /// The completed payload's digest is not what was expected.
     DigestMismatch,
-    /// Try deleting some files!!!
-    SomethingElseWentWrong,
+    /// Something specific to this store implementation went wrong.
+    OperationError(OE),
 }
 
 /// Returned when no entry was found for some criteria.
@@ -112,6 +125,7 @@ where
 {
     type FlushError;
     type BulkIngestionError;
+    type OperationsError;
 
     /// The [namespace](https://willowprotocol.org/specs/data-model/index.html#namespace) which all of this store's [`AuthorisedEntry`] belong to.
     fn namespace_id() -> N;
@@ -125,7 +139,7 @@ where
     ) -> impl Future<
         Output = Result<
             EntryIngestionSuccess<MCL, MCC, MPL, N, S, PD, AT>,
-            EntryIngestionError<MCL, MCC, MPL, N, S, PD, AT>,
+            EntryIngestionError<MCL, MCC, MPL, N, S, PD, AT, Self::OperationsError>,
         >,
     >;
 
@@ -138,8 +152,18 @@ where
         prevent_pruning: bool,
     ) -> impl Future<
         Output = Result<
-            Vec<BulkIngestionResult<MCL, MCC, MPL, N, S, PD, AT>>,
-            BulkIngestionError<MCL, MCC, MPL, N, S, PD, AT, Self::BulkIngestionError>,
+            Vec<BulkIngestionResult<MCL, MCC, MPL, N, S, PD, AT, Self::OperationsError>>,
+            BulkIngestionError<
+                MCL,
+                MCC,
+                MPL,
+                N,
+                S,
+                PD,
+                AT,
+                Self::BulkIngestionError,
+                Self::OperationsError,
+            >,
         >,
     >;
 
@@ -158,7 +182,12 @@ where
         expected_digest: PD,
         expected_size: u64,
         producer: &mut Producer,
-    ) -> impl Future<Output = Result<PayloadAppendSuccess<MCL, MCC, MPL, N, S, PD>, PayloadAppendError>>
+    ) -> impl Future<
+        Output = Result<
+            PayloadAppendSuccess<MCL, MCC, MPL, N, S, PD>,
+            PayloadAppendError<Self::OperationsError>,
+        >,
+    >
     where
         Producer: BulkProducer<Item = u8>;
 
--- a/data-model/src/store.rs
+++ b/data-model/src/store.rs
@@ -201,7 +201,7 @@ where
     ///
     /// If the `traceless` parameter is `true`, the store will keep no record of ever having had the entry. If `false`, it *may* persist what was forgetten for an arbitrary amount of time.
     ///
-    /// Forgetting is not the same as deleting! Subsequent joins with other [`Store`]s may bring the forgotten entry back.
+    /// Forgetting is not the same as [pruning](https://willowprotocol.org/specs/data-model/index.html#prefix_pruning)! Subsequent joins with other [`Store`]s may bring the forgotten entry back.
     fn forget_entry(
         path: &Path<MCL, MCC, MPL>,
         subspace_id: S,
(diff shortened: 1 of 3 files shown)


AljoschaMeyer:
# Design Notes on a Willow Store Trait: Mutation

For our rust implementation of Willow, we are designing a trait to abstract over data stores. Among the features that a store must provide are ingesting new entries, querying areas, resolving payloads, and subscribing to change notifications. Turns out this trait becomes rather involved. In this writeup, I want to focus on a small subset of the trait: everything that allows user code to mutate a data store.

On the surface, Willow stores support only a single mutating operation: ingesting new entries. Does the following trait (heavily simplified on the type-level) do the trick?

```rust
trait Store {
    async fn ingest(&mut self, entry: Entry) -> Result<(), StoreError>;
}
```

Nope, there is actually a whole lot more to consider. We'll start with something simple: while the data model only considers adding new entries to a store, there is a second operation that all implementations should support: locally deleting entries. We want to support this both for explicit entries, and for whole areas:

```rust
trait Store {
    async fn ingest(&mut self, entry: Entry) -> Result<(), StoreError>;
    async fn forget(&mut self, entry: Entry) -> Result<(), StoreError>;
    async fn forget_area(&mut self, area: Area3d) -> Result<(), StoreError>;
    // We also need ingestion and forgetting of payloads,
    // but we'll leave that for later.
}
```

Now we have a sketch of a somewhat useable API, but it does not admit particularly efficient implementations. We want stores to (potentially) be backed by persistent storage. But persisting data to disk is expensive. Typically, writes are performed to an in-memory buffer, and then periodically (or explicitly on demand) flushed to disk (compare the [fsync syscall](https://man7.org/linux/man-pages/man2/fsync.2.html) in posix). To support this pattern, we should change the semantics of our methods to "kindly asking the store to eventually do something", and add a `flush` method to force an expensive commitment of all prior mutations (typically by fsyncing a [write-ahead log](https://en.wikipedia.org/wiki/Write-ahead_logging)).

```rust
trait Store {
    async fn ingest(&mut self, entry: Entry) -> Result<(), StoreError>;
    async fn forget(&mut self, entry: Entry) -> Result<(), StoreError>;
    async fn forget_area(&mut self, area: Area3d) -> Result<(), StoreError>;
    async flush(&mut self) -> Result<(), StoreError>;
}
```

Another typical optimisation is that of efficiently operating on slices of bulk data instead of individual pieces of data. Forgetting should hopefully be rare enough, but we should definitely have a method for bulk ingestion.

```rust
trait Store {
    async fn ingest(&mut self, entry: Entry) -> Result<(), StoreError>;
    async fn ingest_bulk(&mut self, entries: &Entry[]) -> Result<(), StoreError>;
    async fn forget(&mut self, entry: Entry) -> Result<(), StoreError>;
    async fn forget_area(&mut self, area: Area3d) -> Result<(), StoreError>;
    async flush(&mut self) -> Result<(), StoreError>;
}
```

The next issue is one of interior versus exterior mutability. Those async methods desugar to returning Futures. And since the methods take a mutable reference to the store as an argument, no other store methods can be called while such a Future is in scope. Hence, this trait forces sequential issuing of commands, with no concurrency. While it might be borderline acceptable to force linearization of all mutating store accesses (especially since most method calls would simply place an instruction in a buffer rather than actually performing and committing side-effects), the inability to mutate the store while also querying it (say, by subscribing to changes) is a dealbreaker. Thus, we should change those methods to take an immutable `&self` reference, forcing store implementatons to employ interior mutability. To the experienced rust programmer, this shouldn't be too surprising: the whole point of a store is to provide mutable, aliased state, after all.

```rust
trait Store {
    async fn ingest(& self, entry: Entry) -> Result<(), StoreError>;
    async fn ingest_bulk(&self, entries: &Entry[]) -> Result<(), StoreError>;
    async fn forget(&self, entry: Entry) -> Result<(), StoreError>;
    async fn forget_area(&self, area: Area3d) -> Result<(), StoreError>;
    async flush(&self) -> Result<(), StoreError>;
}
```

Finally, there is a line of thought that I'm less confident about yet. Starting with simple `ingest` operations and then adding support for buffering (and flushing of the buffer) and bulk operations duplicates a lot of the design of the ufotofu [`BulkConsumer`](https://docs.rs/ufotofu/latest/ufotofu/local_nb/trait.BulkConsumer.html). So perhaps it would make sense to simply expose a BulkConsumer, whose items are an enum for the different operations (ingest, forget, forget_area). Replacing method calls with explicitly moving around enum values might sound inefficient, but there's an argument to be made that any efficient buffering operation would implement the method calls by storing a reification of the operation inside some buffer anyways. The main upside would be that of using a principled abstraction instead of providing a zoo of methods that end up implementing the same semantics anyways.

Aside from the drawback of forcing the explicit reification of operations, another drawback of the consumer approach would be the question of how to obtain the consumer. If the store itself implemented `BulkConsumer`, then the issue of exerior mutability would make it unusuable. If the store had a function that took a `&self` and returned an owned handle that implements `BulkConsumer`, then arbitrarily many handles could be created, i.e., the consumer would be forced to support many writers. I don't really see a way around that. So the two options seem to either be a multi-writer consumer, or a collection of (interiorly mutable) methods on the store trait without an explicit consumer.

This is the main issue I wanted to convey in this writeup. I have glossed over various details (generics and assocated types, precise errors, information to return on successful operations, an ingest-unless-it-would-overwrite method, payload handling, etc.), because those are comparatively superficial. But I'd love to hear some feedback on the issues of interior vs exterior mutability and explicit vs implicit (buffered, bulk) consumer semantics.

data-model/src/store.rs
```diff
@@ -0,0 +1,104 @@
...
+    NotEntryReference,
+    /// The payload is already held in storage.
+    AlreadyHaveIt,
+    /// The received's payload digest is not what was expected.
+    DigestMismatch,
+    /// Try deleting some files!!!
+    SomethingElseWentWrong,
+}
+
+pub enum QueryOrder {
```
AljoschaMeyer, line 55: Note for later: this is missing an important order: `IDontCareJustDoSomethingEfficientPlease`.

AljoschaMeyer:
# Storage requirements

- user-facing
  - mutation
    - ingest entry
      - optionally do nothing if it would overwrite a (strict) extension of the path
    - append to payload prefix
    - forget entry by pair of path and subspace id
    - forget by area (of interest)
    - forget all *but* area (of interest) (within a containing area (of interest)) (?)
    - all *forget* functionality but for payloads instead of entries
    - all *forget* functionality comes with a *traceless flag*
      - if *traceless*, the storage keeps no record of ever having had the data
      - if not *traceless*, the storage is allowed to persist the *forget* query for an arbitrary amount of time, in order to accurately inform persistent subscribers about the forgetting in the future
        - non-traceless forgetting does *not* imply rejecting future data that matches the *forget* operation, such a service must live at a different level (and must be implemented by not ingesting the data again in the first place)
    - flush (force persistence of all prior mutations)
      - one-shot queries may or may not observe mutations before they were flushed
      - subscriptions are only notified of mutations that have been flushed successfully
  - query
    - query entry by pair of path and subspace id
      - yields Entry, AuthorisationToken, available payload prefix (length)
      - option to filter out results with incomplete payloads
    - query entries by area (of interest)
      - all of the functionality of singleton queries, also
      - ordering results (arbitrary, PTS (optionally reversed), TSP (optionally reversed), SPT (optionally reversed)) (?)
    - subscriptions: all queries should also work as long-lived subscriptions
      - this includes subscription to payload prefix appending
      - this also includes notifications about overwriting and forgetting things
    - persistent subscriptions
      - all subscription notifications come with a u64 *progress id*, client code can stop consuming a subscription at any time and can later (even between program shutdown and restart) resume it at the same point by supplying its progress id
      - this is a best-effort service, the storage may reject a resumption because it is too outdated
      - forgetting entries/payloads can cause progress ids to become outdated, unless the storage tracks the things it forgot
        - this is why all *forget* functionality comes with a flag whether it should be completely traceless or whether the storage is allowed to track the act of forgetting
- internal (for replication)
  - all of the public-facing functionality may also be used internally
  - query entries by 3dRange
    - yields Entry, AuthorisationToken, available payload prefix (length)
    - option to filter out results with incomplete payloads (?)
    - optionally order according to the [`ReconciliationAnnounceEntries:will_sort`](https://willowprotocol.org/specs/sync/index.html#ReconciliationAnnounceEntriesWillSort) field
    - optionally as a subscription (?)
  - summarise 3dRange (count, fingerprint)
  - convert area of interest to 3dRange
  - partition a 3dRange into `k` roughly evenly sized 3dRanges

sgwilym:
I'm starting to work in earnest on this again.

One thing I have thought about is how what a store is in our implementations is not what a store is in the spec:

> A [store](https://willowprotocol.org/specs/data-model/index.html#store) is a set of [AuthorisedEntries](https://willowprotocol.org/specs/data-model/index.html#AuthorisedEntry) ...

Whereas in our implementations, a store is a set of authorised entries AND a partial set of corresponding payloads.

In willow-js, this led to this situation where you need to specify how a store's payloads should be stored and retrieved via PayloadDrivers, ramping up the complexity of the store.

I suggested the opportunity to separate the two concepts to @AljoschaMeyer, so that we'd have `Store`s concerned only with entries and, err, `PayloadStore`s concerned only with Payloads. Rather than have `Store.ingest_payload`, we could have `Store.update_available_payload_length` to update the fingerprints.

But Aljoscha raised a good point:

> I feel like internally the implementations need to be coupled though, say, for locking.

Which is persuasive, but maybe there's some way around it? Which is why I'm sharing this here.

sgwilym:
@AljoschaMeyer I have now added all mutation methods (unless we want a bulk payload ingestion method?)

data-model/src/store.rs
```diff
@@ -0,0 +1,228 @@
...
+    Obsolete {
+        /// The obsolete entry which was not ingested.
+        obsolete: AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>,
+        /// The newer entry which was not overwritten.
+        newer: AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>,
+    },
+}
+
+/// Returned when an entry cannot be ingested into a [`Store`].
+pub enum EntryIngestionError<
```
Frando, line 36: This and most other errors should include a variant `Io(Self::IoError)` or so I think, because persistent store operations are fallible by nature and the error has to bubble up.

data-model/src/store.rs
```diff
@@ -0,0 +1,228 @@
...
+    PD: PayloadDigest,
+{
+    /// The payload was appended to but not completed.
+    Appended(Vec<LengthyEntry<MCL, MCC, MPL, N, S, PD>>),
+    /// The payload was completed by the appendment.
+    Completed(Vec<LengthyEntry<MCL, MCC, MPL, N, S, PD>>),
+}
+
+/// Returned when a payload fails to be appended into the [`Store`].
+pub enum PayloadAppendError {
```
Frando, line 89: Needs `IoError(Self::IoError)` variant I think

data-model/src/store.rs
```diff
@@ -0,0 +1,228 @@
...
+        traceless: bool,
+    ) -> impl Future<Output = Vec<AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>>>;
+
+    /// Locally forget a payload with a given [`PayloadDigest`], or an error if no payload with that digest is held by this store.
+    ///
+    /// If the `traceless` parameter is `true`, the store will keep no record of ever having had the payload. If `false`, it *may* persist what was forgetten for an arbitrary amount of time.
+    ///
+    /// Forgetting is not the same as deleting! Subsequent joins with other [`Store`]s may bring the forgotten payload back.
+    fn forget_payload(
+        digest: PD,
```
Frando, line 202: Shouldn't this get an `Entry` instead of a `digest` as arguments? Otherwise apps would have to do their own reference counting. If this takes an `Entry`, the payload would be deleted only if it is not referenced by other entries?

data-model/src/store.rs
```diff
@@ -0,0 +1,228 @@
...
+    const MPL: usize,
+    N: NamespaceId,
+    S: SubspaceId,
+    PD: PayloadDigest,
+    AT: AuthorisationToken<MCL, MCC, MPL, N, S, PD>,
+> {
+    /// The entry was successfully ingested.
+    Success,
+    /// The entry was successfully ingested and prefix pruned some entries.
+    SuccessAndPruned(Vec<AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>>),
```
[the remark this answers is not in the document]
AljoschaMeyer: This is a great example of why standard libraries should have a notion of the opposite of an iterator (i.e., consumers). The functions can take a mutable reference to a consumer, and feed pruned entries into that consumer.

data-model/src/store.rs
AljoschaMeyer: I very much like `Self::IoError` over `std::io::Error`. Perhaps it shouldn't even be called `IoError`, but have a more generic name? `OperationsError`?

data-model/src/store.rs
AljoschaMeyer: Which is a slightly more principled solution than the `SomethingElseWentWrong` variant, I suppose...

data-model/src/store.rs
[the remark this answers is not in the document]
Frando: Oh, sure, that is much better than returning a producer.

data-model/src/store.rs
```diff
@@ -0,0 +1,228 @@
...
+        Output = Result<
+            EntryIngestionSuccess<MCL, MCC, MPL, N, S, PD, AT>,
+            EntryIngestionError<MCL, MCC, MPL, N, S, PD, AT>,
+        >,
+    >;
+
+    /// Attempt to ingest many [`AuthorisedEntry`] in the [`Store`].
+    ///
+    /// The result being `Ok` does **not** indicate that all entry ingestions were successful, only that each entry had an ingestion attempt, some of which *may* have returned [`EntryIngestionError`]. The `Err` type of this result is only returned if there was some internal error.
+    fn bulk_ingest_entry(
```
AljoschaMeyer, line 135: Returning a Vec of errors causes an allocation that might be avoidable (or at least configurable) if the method took a consumer of BulkIngestionResults as an argument instead? Not entirely sure whether that would really be an improvement. CC @Frando

data-model/src/store.rs
```diff
@@ -0,0 +1,228 @@
...
+            BulkIngestionError<MCL, MCC, MPL, N, S, PD, AT, Self::BulkIngestionError>,
+        >,
+    >;
+
+    /// Attempt to append part of a payload for a given [`AuthorisedEntry`].
+    ///
+    /// Will fail if:
+    /// - The payload digest is not referred to by any of the store's entries.
+    /// - A complete payload with the same digest is already held in storage.
+    /// - The payload exceeded the expected size
```
AljoschaMeyer, line 151: Isn't "A complete payload with the same digest is already held in storage." just a special case of "The payload exceeded the expected size"?

data-model/src/store.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// - The final payload's digest did not match the expected digest
+    /// - Something else went wrong, e.g. there was no space for the payload on disk.
+    ///
+    /// This method **cannot** verify the integrity of partial payload. This means that arbitrary (and possibly malicious) payloads smaller than the expected size will be stored unless partial verification is implemented upstream (e.g. during [the Willow General Sync Protocol's payload transformation](https://willowprotocol.org/specs/sync/index.html#sync_payloads_transform)).
```
AljoschaMeyer, line 155:
```suggestion
    /// This method **cannot** verify the integrity of partial payloads. This means that arbitrary (and possibly malicious) payloads smaller than the expected size will be stored unless partial verification is implemented upstream (e.g. during [the Willow General Sync Protocol's payload transformation](https://willowprotocol.org/specs/sync/index.html#sync_payloads_transform)).
```

data-model/src/store.rs
```diff
@@ -0,0 +1,228 @@
...
+    fn append_payload<Producer>(
+        &self,
+        expected_digest: PD,
+        expected_size: u64,
+        producer: &mut Producer,
```
AljoschaMeyer, line 160: Don't love this argument name. If it stays like this, the docs should at least state that the `producer` is how the payload data is being supplied into the store.

data-model/src/store.rs
```diff
@@ -0,0 +1,228 @@
...
+    ) -> impl Future<Output = Result<PayloadAppendSuccess<MCL, MCC, MPL, N, S, PD>, PayloadAppendError>>
+    where
+        Producer: BulkProducer<Item = u8>;
+
+    /// Locally forget an entry with a given [path] and [subspace] id, returning the forgotten entry, or an error if no entry with that path and subspace ID are held by this store.
+    ///
+    /// If the `traceless` parameter is `true`, the store will keep no record of ever having had the entry. If `false`, it *may* persist what was forgetten for an arbitrary amount of time.
```
AljoschaMeyer, line 167:
```suggestion
    /// If the `traceless` parameter is `true`, the store will keep no record of ever having had the entry. If `false`, it *may* persist what was forgotten for an arbitrary amount of time.
```

data-model/src/store.rs
```diff
@@ -0,0 +1,228 @@
...
+    ///
+    /// Forgetting is not the same as deleting! Subsequent joins with other [`Store`]s may bring the forgotten entry back.
```
AljoschaMeyer, line 169: Given that willow has no notion of "deleting", this sentence might be confusing/misleading.

data-model/src/store.rs
```diff
@@ -0,0 +1,228 @@
...
+    fn forget_entry(
+        path: &Path<MCL, MCC, MPL>,
+        subspace_id: S,
+        traceless: bool,
+    ) -> impl Future<Output = Result<AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>, NoSuchEntryError>>;
```
AljoschaMeyer, line 174: A part of me wants this method to return `Result<(), Self::OperationsError>`, but nothing more. Report whether there might still be an entry (the error case), or whether there definitely is no entry anymore (the ok case), irrespective of whether there was one before, or what it might have been.

The forgetting method copying the entry into memory and presenting it seems... odd?

data-model/src/store.rs
```diff
@@ -0,0 +1,228 @@
...
+        traceless: bool,
+    ) -> impl Future<Output = Vec<AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>>>;
+
+    /// Locally forget all [`AuthorisedEntry`] **not** [included](https://willowprotocol.org/specs/grouping-entries/index.html#area_include) by a given [`AreaOfInterest`], returning all forgotten entries
+    ///
+    /// If the `traceless` parameter is `true`, the store will keep no record of ever having had the forgotten entries. If `false`, it *may* persist what was forgetten for an arbitrary amount of time.
+    ///
+    /// Forgetting is not the same as deleting! Subsequent joins with other [`Store`]s may bring the forgotten entries back.
+    fn forget_everything_but_area(
+        area: &AreaOfInterest<MCL, MCC, MPL, S>,
```
AljoschaMeyer, line 192: Think this could use a `container` parameter. You give it an outer `container` area, outside of which nothing will be forgotten. Inside it, everything is forgotten but the `inner` area. Setting the container to the full area recovers the current functionality.

data-model/src/store.rs
AljoschaMeyer: Agreed. I think willow-rs clients should only interface with the db in terms of entries, not in terms of payloads.

data-model/src/store.rs
sgwilym: Basically I am trying to anticipate the situation where the producer keeps producing bytes and we know that it is going to be the wrong payload because we know the expected size.

Or will there be producer re-use between payload ingestions?

data-model/src/store.rs
sgwilym: I find this confusing. You could also have a `inner` parameter on `forget_area`, too, which would be functionally equivalent.

data-model/src/store.rs
AljoschaMeyer: I see. Makes sense, but could use some clarification then.

data-model/src/store.rs
AljoschaMeyer: Fair, either works.

data-model/src/store.rs
sgwilym: I mean, I think you would always want to know about errors, right? The consumer pattern is cool, but is it maybe a little fancy-shmancy? 😛

data-model/src/store.rs
```diff
@@ -0,0 +1,380 @@
...
+/// Describes an [`AuthorisedEntry`] which was pruned and the [`AuthorisedEntry`] which triggered the pruning.
+pub struct PruneEvent<const MCL: usize, const MCC: usize, const MPL: usize, N, S, PD, AT>
+where
+    N: NamespaceId,
+    S: SubspaceId,
+    PD: PayloadDigest,
+    AT: AuthorisationToken<MCL, MCC, MPL, N, S, PD>,
+{
+    /// The entry which was pruned.
+    pub pruned: AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>,
```
AljoschaMeyer, line 137: In the interest of minimising keeping around information about pruned things, this could be a pair of a subspaceid and a path instead of the full entry information.

I'm on the fence about whether programmers should be given the full information, or whether the API should ooze commitment to deletion.

data-model/src/store.rs
```diff
@@ -0,0 +1,380 @@
...
+    S: SubspaceId,
+    PD: PayloadDigest,
+    AT: AuthorisationToken<MCL, MCC, MPL, N, S, PD>,
+{
+    /// A new entry was ingested.
+    Ingested(u64, AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>),
+    /// An existing entry received a portion of its corresponding payload.
+    Appended(u64, LengthyAuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>),
+    /// An entry was forgotten.
+    EntryForgotten(u64, AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>),
```
AljoschaMeyer, line 156: Here again, the bare minimum would be a subspaceid, path pair.

AljoschaMeyer: Compare to PayloadForgotten, we'd never include the payload itself in that.

data-model/src/store.rs
AljoschaMeyer: This is philosophical enough to also ping @Frando probalby?

data-model/src/store.rs
```diff
@@ -0,0 +1,380 @@
...
+        subspace_id: S,
+        traceless: bool,
+    ) -> impl Future<Output = Result<(), ForgetPayloadError>>;
+
+    /// Locally forget the corresponding payload of the entry with a given path and subspace, or an error if no entry with that path and subspace ID is held by this store. **The payload will be forgotten even if it corresponds to other entries**.
+    ///
+    /// If the `traceless` parameter is `true`, the store will keep no record of ever having had the payload. If `false`, it *may* persist what was forgetten for an arbitrary amount of time.
+    ///
+    /// Forgetting is not the same as [pruning](https://willowprotocol.org/specs/data-model/index.html#prefix_pruning)! Subsequent joins with other [`Store`]s may bring the forgotten payload back.
+    fn forget_payload_unchecked(
```
AljoschaMeyer, line 309: Bikeshedding, but "unchecked" makes me think of unsafe code or invariants of types. How about "force_forget_payload"?

AljoschaMeyer: Also, this makes me wonder why the bulk forget operations don't also have a variant like this.

data-model/src/store.rs
sgwilym: I see what you mean. I think what I was trying to anticipate was giving application authors extra information they might need to remove things from the UI without needing to e.g. refetch an entire area.

data-model/src/store.rs
```diff
@@ -0,0 +1,380 @@
...
+    }
+
+    pub fn ignore_empty_payloads(&mut self) {
+        self.ignore_empty_payloads = true;
+    }
+}
+
+/// Returned when a payload could not be forgotten.
+pub enum ForgetPayloadError {
+    NoSuchEntry,
```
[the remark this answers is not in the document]
sgwilym: I feel like there may be scenarios you'd want to know to distinguish whether something was forgotten or whether it was never there at all.

data-model/src/store.rs
AljoschaMeyer: Yeah. The  main question is whether we should force developers to ensure that subspace and path are sufficient to do all their cleanup. And I'm inclined to say that this is indeed the way.
We could always add this later, if people ran into an actual wall.

data-model/src/store.rs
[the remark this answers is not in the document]
AljoschaMeyer: Yeah, fair.

data-model/src/store.rs
```diff
@@ -0,0 +1,386 @@
...
+        ignore: Option<QueryIgnoreParams>,
+    ) -> impl Producer<Item = LengthyAuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>>;
+
+    /// Subscribe to events concerning entries [included](https://willowprotocol.org/specs/grouping-entries/index.html#area_include) by an [`AreaOfInterest`], returning a producer of `StoreEvent`s which occurred since the moment of calling this function.
+    ///
+    /// If `ignore_incomplete_payloads` is `true`, the producer will not produce entries with incomplete corresponding payloads.
+    /// If `ignore_empty_payloads` is `true`, the producer will not produce entries with a `payload_length` of `0`.
+    fn subscribe_area(
+        &self,
+        area: &AreaOfInterest<MCL, MCC, MPL, S>,
```
Frando, line 375: Shouldn't those take an `Area` and not an `AreaOfInterest`? What would max_count and max_size mean in the context of a subscription?

sgwilym: Agreed, thank you!

Frando:
Note: Many of the structs still need the standard derives, Debug, Clone, and Eq/PartialEq where applicable.

data-model/src/store.rs
```diff
@@ -188,6 +192,14 @@ pub enum ForgetPayloadError {
     ReferredToByOtherEntries,
 }
 
+/// The origin of an entry ingestion event.
+pub enum EntryOrigin {
+    /// The entry was probably created on this machine.
+    Local,
+    /// The entry was sourced from another device, e.g. a networked sync session.
```
AljoschaMeyer, line 350: What is the `u64` in this variant? The docs should say.

PR #22, opened 2024-07-01
Path Successors [✨ surprises ✨]

sgwilym:
Last week I naïvely thought: "let's quickly port the willow-utils path successor functions and get that out of the way". Things did not work out that way!

**TL;DR**: _All_ path successor implementations up to this point are wrong, and we were only able to find out how using fuzz testing. willow-util's implementations need to be fixed, as it may be generating the wrong ranges for areas.

## `SubspaceId`

- Added `successor` method.

## `PathComponent` trait

- Added `new_with_tail` method
- Added `empty` method
- Added `try_append_zero_byte` method
- Added `set_byte` method
- Added `try_increment_fixed_width` method
- Added `prefix_successor` method

## `Path` trait

- The iterator returned by the `components` method must implement `ExactSizeIterator` and `DoubleEndedIterator`
- Added `component_count` method
- Added `is_empty` method
- Fixed a bug `is_prefix_of`
- Added `successor` method
- Added `successor_of_prefix` method
- Added explicit implementations of `Ord`, `PartialOrd`

## Testing

- Implemented `Arbitrary` for `PathComponentBox`
- Implemented `Arbitrary` for `PathRc`
- Added fuzz tests for `Path::successor`
- Added fuzz tests for `Path::successor_of_prefix`


sgwilym:
@Frando you may want to check this against your own implementation for path successors.

Issue #17, opened 2024-06-20
Willow parameters for testing?


sgwilym:
From here on out, testing our modules will require an increasing number of Willow parameters to be configured.

In willow-js, I made a kind of test scheme with extremely simplified parameters: `NamespaceId` was an enum of `Project`, `Work`, `Family`; `SubspaceId` was an enum of `Alfie`, `Betty`, `Gemma` etc. The sometimes tricky aspect of this was making sure that these simple parameters weren't _too_ simple and actually tested the properties of Willow. And some end _can't_ be simple, like the parameters for private area intersection.

So I suggest there's two ways we can go with this: either devise a minimally viable set of parameters for use in testing, or use a real set of parameters, e.g. Earthstar's.

mycognosist:
We could probably benefit from using [rstest](https://docs.rs/rstest/latest/rstest/) in these cases. It provides a neat means of creating test fixtures and parametrised tests. There are lots of good usage examples in the p2panda codebase 🐼

AljoschaMeyer:
Not sure if rstest helps too much with supplying choices type parameters ("generics"), though.

sgwilym:
> We could probably benefit from using [rstest](https://docs.rs/rstest/latest/rstest/) in these cases. It provides a neat means of creating test fixtures and parametrised tests. There are lots of good usage examples in the p2panda codebase 🐼

This question isn't so much about setting up specific states with reusable items (e.g. give me a `Store` with all the entries for a chat between Alfie and Betty in it), but about something lower than that: in order to test any fixtures _at all_ we need to 
decide which parameters Willow uses (e.g. that a `SubspaceId` _can be_ `Alfie` or `Betty`). These parameters need to be 'real' enough to adequately test Willow's qualities. For instance, whatever we use for the `is_authorised_write` parameter needs to succeed or fail based on the entry used and the token given and can't just be `() => true`.

So my question is: do we do what willow-js does (fantasy protocol with reasonable constraints) or just use Earthstar's parameters, which are all defined already?

- Using fantasy parameters is faster, but the expediency may lead to blind spots in our tests.
- Using Earthstar's parameters would be some more upfront work (e.g. implementing `Cinn25519`), but the tests could give us good confidence, and we'd kind of implement Earthstar along the way.

AljoschaMeyer:
As someone who hopefully won't need to implement any of it, I'd prefer using the Earthstar parameters directly =P

sgwilym:
[...]

PR #23, opened 2024-07-09
Encoding traits, Path + Entry impls

sgwilym:
- Add a `compact_width` module
- Add a `max_power` module
- Adds `Encoder` and `Decoder` traits which work with Ufotofu consumers and producers. 
- Implements `Encoder` + `Decoder` on PathRc
- Implements `Encoder` + `Decoder` on Entry
- Adds an `earthstar` crate with nascent `NamespaceId` and `IdentityId` structs
  - These are built on top of a wip Cinn25519 implementation with validated shortnames.
  - I did this to implement `Arbitrary` on these types so we could use realistic values in fuzz tests as per #17.

This is already enough to review, I will do the relative encodings in a separate branch off of this.

Still needs:

- [ ] A non-local Ufotofu dependency


data-model/src/encoding/compact_width.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// The byte-width required to represent numbers up to 256^2 (i.e. a 16-bit number).
+    Two,
+    /// The byte-width required to represent numbers up to 256^4 (i.e. a 32-bit number).
+    Four,
+    /// The byte-width required to represent numbers up to 256^8 (i.e. a 64-bit number).
+    Eight,
+}
+
+#[derive(Debug)]
+pub struct NotACompactWidthError(pub u8);
```
AljoschaMeyer, line 18: Feels like an internal error that should not be exposed to me. Have not looked at where it is used yet, though.

data-model/src/encoding/compact_width.rs
```diff
@@ -0,0 +1,171 @@
...
+    };
+
+    consumer
+        .bulk_consume_full_slice(&value.to_be_bytes()[8 - width..])
+        .await?;
+
+    Ok(())
+}
+
+pub async fn decode_u8_be<Producer: BulkProducer<Item = u8>>(
```
AljoschaMeyer, line 69: Bit surprised to find this family of functions in a file named `compact_width.rs`, given the wide applicability elsewhere.

I could see us moving these into a dedicated crate even. Compare the `put_*` methods on https://docs.rs/bytes/1.6.0/bytes/buf/trait.BufMut.html#provided-methods

data-model/src/encoding/compact_width.rs
```diff
@@ -0,0 +1,171 @@
...
+    producer: &mut Producer,
+) -> Result<u8, DecodeError<Producer::Error>> {
+    let mut byte = [0; 1];
```
AljoschaMeyer, line 72: Even more clean would be `size_of::<u8>` instead of  `1` as the array size.

data-model/src/encoding/error.rs
```diff
@@ -0,0 +1,59 @@
+use either::Either;
+use std::num::TryFromIntError;
+use ufotofu::{
+    common::errors::{ConsumeFullSliceError, OverwriteFullSliceError},
+    local_nb::producer::PipeIntoSliceError,
+};
+
+/// Returned when a encoding fails to be consumed by a [`ufotofu::local_nb::Consumer`].
+#[derive(Debug)]
```
AljoschaMeyer, line 9: Needs an impl of the `Error` trait; can also derive `Clone, PartialEq, Eq` at the very least.

data-model/src/encoding/parameters.rs
```diff
@@ -0,0 +1,26 @@
+use std::future::Future;
+
+use crate::encoding::error::{DecodeError, EncodingConsumerError};
+use ufotofu::local_nb::{BulkConsumer, BulkProducer};
+
+pub trait Encoder<Consumer>
```
AljoschaMeyer, line 6: The consumer should not be a parameter of the trait, but just of the `encode` function. That would prevent implementations of `Encoder` from relying on any details of the BulkConsumer beyond the BulkConsumer trait.

data-model/src/encoding/parameters.rs
```diff
@@ -0,0 +1,26 @@
...
+    Consumer: BulkConsumer<Item = u8>,
+{
+    /// A function from the set `Self` to the set of bytestrings, ensuring that any value of `Self` maps to exactly one bytestring.
+    fn encode(
+        &self,
+        consumer: &mut Consumer,
+    ) -> impl Future<Output = Result<(), EncodingConsumerError<Consumer::Error>>>;
+}
+
+pub trait Decoder<Producer>
```
AljoschaMeyer, line 17: See above.

data-model/src/encoding/parameters.rs
```diff
@@ -0,0 +1,26 @@
+use std::future::Future;
+
+use crate::encoding::error::{DecodeError, EncodingConsumerError};
+use ufotofu::local_nb::{BulkConsumer, BulkProducer};
+
+pub trait Encoder<Consumer>
+where
+    Consumer: BulkConsumer<Item = u8>,
+{
+    /// A function from the set `Self` to the set of bytestrings, ensuring that any value of `Self` maps to exactly one bytestring.
```
AljoschaMeyer, line 10: Can link to the spec imo. The requirements are stricter than just this doc comment (mapping every possible value to the empty string currently satisfies the doc comment).

data-model/src/entry.rs
```diff
@@ -58,6 +66,101 @@ where
...
+
+        consumer
+            .bulk_consume_full_slice(&self.payload_length.to_be_bytes())
+            .await?;
+
+        self.payload_digest.encode(consumer).await?;
+
+        Ok(())
+    }
+}
```
AljoschaMeyer, line 94: So nice =)

data-model/src/entry.rs
```diff
@@ -58,6 +66,101 @@ where
...
+    P: Path + Decoder<Prod>,
+    PD: PayloadDigest + Decoder<Prod>,
+    Prod: BulkProducer<Item = u8>,
+{
+    async fn decode(producer: &mut Prod) -> Result<Self, DecodeError<Prod::Error>> {
+        let namespace_id = N::decode(producer).await?;
+        let subspace_id = S::decode(producer).await?;
+        let path = P::decode(producer).await?;
+
+        let mut timestamp_bytes = [0u8; 8];
```
AljoschaMeyer, line 109: Again, `size_of::<u64>()` instead of `8` would be more explicit.

data-model/src/entry.rs
```diff
@@ -58,6 +66,101 @@ where
...
+        producer
+            .bulk_overwrite_full_slice(&mut timestamp_bytes)
+            .await?;
+        let timestamp = u64::from_be_bytes(timestamp_bytes);
+
+        let mut payload_length_bytes = [0u8; 8];
```
AljoschaMeyer, line 115: I smell copy-pasting that could be reduced to a one-liner function call here.

We *could* implement `Encoder` and `Decoder` for `u8, u16, u32, u64` directly, using a big-endian encoding. Or we can create thin wrapper types `U64BE, U32BE, ...` that implement those traits and `From<u64>, From<u32>, ...`.

Then, in this function, we'd have

```rust
let timestamp: u64 = U64BE::decode(producer).await?.into();
let payload_length: u64 = U64BE::decode(producer).await?.into();
```

AljoschaMeyer: With such types, the *en*coding of entries (and all future types) could also become more uniform.

AljoschaMeyer: We should do a call on this I think.

data-model/src/entry.rs
```diff
@@ -58,6 +66,101 @@ where
...
+            subspace_id,
+            path,
+            timestamp,
+            payload_length,
+            payload_digest,
+        })
+    }
+}
+
+impl<'a, N, S, P, PD> Arbitrary<'a> for Entry<N, S, P, PD>
```
AljoschaMeyer, line 134: Should become feature-gated. Also a topic to discuss synchronously on a call.

data-model/src/entry.rs
```diff
@@ -58,6 +66,101 @@ where
...
+            subspace_id: Arbitrary::arbitrary(u)?,
+            path: Arbitrary::arbitrary(u)?,
+            payload_digest: Arbitrary::arbitrary(u)?,
+            payload_length: Arbitrary::arbitrary(u)?,
+            timestamp: Arbitrary::arbitrary(u)?,
+        })
+    }
+
+    fn size_hint(depth: usize) -> (usize, Option<usize>) {
+        and_all(&[
```
AljoschaMeyer, line 153: Ooh, didn't know about this one.

data-model/src/parameters.rs
```diff
@@ -21,6 +21,7 @@ pub trait SubspaceId: Ord + Default + Clone {
 /// [Definition](https://willowprotocol.org/specs/data-model/index.html#PayloadDigest).
 pub trait PayloadDigest: Ord + Default + Clone {}
 
+// FIXME: This type is not implementable! Argh!
```
AljoschaMeyer, line 24: Bleurgh!

AljoschaMeyer: (technically it is a *trait* that is not implementable though =P)

data-model/src/path.rs
```diff
@@ -1,5 +1,12 @@
 use arbitrary::{Arbitrary, Error as ArbitraryError, Unstructured};
 use std::rc::Rc;
+use ufotofu::local_nb::{BulkConsumer, BulkProducer};
```
AljoschaMeyer, line 3: Mostly skipping this file for now.

data-model/src/path.rs
```diff
@@ -392,6 +406,83 @@ impl<const MCL: usize, const MCC: usize, const MPL: usize> PartialOrd for PathRc
...
 
+impl<const MCL: usize, const MCC: usize, const MPL: usize, C> Encoder<C> for PathRc<MCL, MCC, MPL>
+where
+    C: BulkConsumer<Item = u8>,
+{
+    async fn encode(&self, consumer: &mut C) -> Result<(), EncodingConsumerError<C::Error>> {
+        let path_length_power = max_power(MCL);
+        let path_count_power = max_power(MCC);
+
+        let path_count_raw: [u8; 8] = self.component_count().to_be_bytes();
```
AljoschaMeyer, line 417: size_of yadda yadda

data-model/src/path.rs
```diff
@@ -392,6 +406,83 @@ impl<const MCL: usize, const MCC: usize, const MPL: usize> PartialOrd for PathRc
...
+
+        consumer
+            .bulk_consume_full_slice(&path_count_raw[8 - (path_count_power as usize)..])
```
AljoschaMeyer, line 420: This looks arcane enough to probably warrant a dedicated, documented function.

data-model/src/path.rs
```diff
@@ -392,6 +406,83 @@ impl<const MCL: usize, const MCC: usize, const MPL: usize> PartialOrd for PathRc
...
+            .await?;
+
+        for component in self.components() {
+            let component_length_raw = component.len().to_be_bytes();
+
+            consumer
+                .bulk_consume_full_slice(&component_length_raw[8 - (path_length_power as usize)..])
+                .await?;
+
+            if component.len() > 0 {
```
AljoschaMeyer, line 430: Why this check?

fuzz/fuzz_targets/entry_encoding.rs
```diff
@@ -0,0 +1,76 @@
+#![no_main]
```
AljoschaMeyer, line 1: How generic can we make the "generate arbitrary instance, encode, decode, check for equality" tests? Can we turn all of this into a generic function that we reuse everywhere? That would reduce the friction for adding such tests for, e.g., fixed-width big-endian integers as well.

fuzz/fuzz_targets/entry_encoding_random.rs
```diff
@@ -0,0 +1,85 @@
+#![no_main]
```
AljoschaMeyer, line 1: Again, could this become fully generic over the type under test?

data-model/src/encoding/compact_width.rs
sgwilym: When I try this I notice two things:

1. It involves importing something from std
2. I get the error "use of unstable library feature 'core_intrinsics': intrinsics are unlikely to ever be stabilized, instead they should be used through stabilized interfaces in the rest of the standard library"

Should I still use this?

data-model/src/encoding/compact_width.rs
sgwilym: If I do not expose it, I am warned that the error is more private than the `new` fn which returns it. The only alternative I know of is to panic, which does not seem good.

data-model/src/encoding/error.rs
sgwilym: This seems like another one for a call.

data-model/src/encoding/max_power.rs
```diff
@@ -0,0 +1,22 @@
+/// Return the least natural number such that 256^`n` is greater than or equal to `n`.
+///
+/// Used for determining the minimal number of bytes needed to represent a given unsigned integer, and more specifically [`path_length_power`](https://willowprotocol.org/specs/encodings/index.html#path_length_power) and [`path_count_power`](https://willowprotocol.org/specs/encodings/index.html#path_count_power).
+pub const fn max_power(n: usize) -> u8 {
```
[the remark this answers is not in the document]
sgwilym: I think my only reasoning here is that compact widths are used quite widely throughout encodings, so it makes sense to put some effort into their API, whereas max_power is exclusively used for path encoding.

data-model/src/encoding/compact_width.rs
AljoschaMeyer: But that `new` function is only used within the crate, not within the public API of the crate, right? `pub(crate)` for the rescue: https://doc.rust-lang.org/reference/visibility-and-privacy.html#pubin-path-pubcrate-pubsuper-and-pubself

data-model/src/encoding/compact_width.rs
AljoschaMeyer: Re 1: import `core::mem::size_of` instead. Always favour importing from `core` rather than `std` if the item exists in `core`.
Re 2: huh, that comes as a surprise to me. What exactly did you import? Does this persist when you import `core::mem::size_of`?

data-model/src/path.rs
AljoschaMeyer: Yes, and I'm not even convinced that this *would* be more efficient.

data-model/src/encoding/compact_width.rs
```diff
@@ -0,0 +1,205 @@
...
+        match self {
+            CompactWidth::One => 1,
+            CompactWidth::Two => 2,
+            CompactWidth::Four => 4,
+            CompactWidth::Eight => 8,
+        }
+    }
+}
+
+impl From<u64> for CompactWidth {
```
AljoschaMeyer, line 45: These `From` implementations will technically work, but they do not match the [intended semantics](https://doc.rust-lang.org/std/convert/trait.From.html). To quote the docs, the conversion "should typically" fulfill:

> The conversion is lossless: semantically, it should not lose or discard information. For example, i32: From<u16> exists, where the original value can be recovered using u16: TryFrom<i32>. And String: From<&str> exists, where you can get something equivalent to the original value via Deref. But From cannot be used to convert from u32 to u16, since that cannot succeed in a lossless way.

data-model/src/encoding/unsigned_int.rs
```diff
@@ -0,0 +1,168 @@
+use ufotofu::local_nb::{BulkConsumer, BulkProducer};
+
+use crate::encoding::error::{DecodeError, EncodingConsumerError};
+use crate::encoding::parameters::{Decoder, Encoder};
+
+/// A big-endian encoded 8-bit unsigned integer.
```
AljoschaMeyer, line 6:
```suggestion
/// A big-endian encoded 8-bit unsigned integer.
///
/// This is the same as a little-endian 8-bit unsigned integer, because endianess only becomes relevant once there is more than one byte. We chose this name for consistency with the other types that we define.
```

AljoschaMeyer: The comment still sounds like the representation of the value in memory was big-endian, but this wrapper type is only relevant for the Encoding and Decoding traits. I think it makes sense to document it explicitly like that. E.g.

```
A `u8` wrapper that implements [`Encoding`] and [`Decoding`] by encoding as a big-endian fixed-width integer.
```

data-model/src/encoding/compact_width.rs
sgwilym: I've removed the `From` impls, but have had trouble coming up with better method names than `from_u8`, `from_u64`, etc.

data-model/src/encoding/compact_width.rs
sgwilym: Using `core::mem::size_of` worked without the warnings 👍

data-model/src/parameters.rs
sgwilym: Bonus: refactored into something people can use.

data-model/src/encoding/compact_width.rs
AljoschaMeyer: Those are fine I think.

AljoschaMeyer:
@Frando 

> The cost is, I guess, that I have to calculate the message length in advance.

To me, the true cost is the write buffer itself. You need to allocate it, and you need to copy its contents into the channel primitive. The ufotofu design allows for zero-copy serialisation, where you write directly to the channel (even if its internal buffer is smaller than the total size of the encoding) without intermediate allocations.

This is a win, and the resulting code is *also* more (or, at least, quite) elegant ([for example, the entry encoder](https://github.com/earthstar-project/willow-rs/blob/encodings/data-model/src/entry.rs#L79)).

The big drawback is having to adopt ufotofu, of course. For obvious reasons I don't consider that as a drawback conceptually, but I cannot talk away the real engineering cost (and risk?), especially since the ufotofu codebase is immature and still undergoing some changes. In particular, you'd need to code against the [`slice_helpers`](https://github.com/worm-blossom/ufotofu/tree/slice_helpers) branch, which should more accurately be called `refactor_all_the_things`.

I'd be up for a call to help with ufotofu integration and just general high-bandwidth sharing.

Frando:
> You need to allocate it, and you need to copy its contents into the channel primitive.

Not necessarily. My encoder trait encodes into `&mut [u8]` with the invariant that the slice is at least `self.encoded_len()` bytes long.

My `Channel::send` takes `item: impl Encoder`, checks if the available buffer is at least `item.encoded_len`, yields if not, and then calls `item.encode` with a mut slice of the channel buffer.

It does not use MaybeUninitialized, but that is an ortogonal concern.

I do like the design of the ufotofu traits, I am just unclear if the coupling to these new and still unstable traits is required for the quite low-level encoder trait.

AljoschaMeyer:
> My encoder trait encodes into &mut [u8] with the invariant that the slice is at least self.encoded_len() bytes long.

I'm not sure whether we are talking past each other. I'm basically interested in the setting where the buffer of the channel (say, the OS-managed buffer for to write tcp data into) is smaller than the encoding. As far as I'm understanding the sentence I quoted, you'd require a channel that can allocate arbitrarily large buffers on demand, which is what the ufotofu design avoids.

Then again, the code you linked uses `io::Write` and not `&mut [u8]`, and `io::Write` does not have this problem. Which is why I feel like I'm misunderstanding something.

In any case, the ufotofu design is pretty much the same as that of `io::Write`, except it fixes a couple of glaring issues (hardcoded item and error type (the latter being important for encoding), inability to pipe from reader to writer without an intermediate buffer, weirdness around uninitialised memory).

> It does not use MaybeUninitialized, but that is an ortogonal concern.

Agreed. I keep switching between being happy or unhappy with the ufotofu approach. But irrespective of those feelings, it will stay the way it is, because it should have the ability to work with uninitialised memory. I just wish it was less of a pain...

> interested if the encoding can be made less coupled to the io traits without perf or ergonomic downsides

Yeah, I share the sentiment, but I didn't find a satisfying way to do it. I think the ability for the channel to dictate fragment sizes is crucial, and that is what the "io<-like> traits" are all about. `Buf` and `BufMut` from `bytes` are surprisingly clsoe in some respects, but they have no error handling...

In my opinion, there is simply a void in terms of popular traits/apis in the rust ecosystem for general, io-like traits. And ufotofu fills that void in a - to me - satisfying manner. It just isn't a *popular* API (yet?).

---

Feel free to reach out on discord if you want to schedule a call.

sgwilym:
Now with zero-copy Path changes merged in.

AljoschaMeyer:
@Frando Just published ufotofu version 0.2.0 fyi, that one should be fairly stable to code against.

PR #24, opened 2024-07-14
Relative encodings

sgwilym:
## New relative encoding traits

Adds `RelativeEncodable` and `RelativeDecodable` traits. But wait! Why the new traits, you ask? Why not just define a `Relativity` struct and implement `Encoder` and `Decoder` on those?

- `Decoder::decode` is a static method which takes a Ufotofu producer as its single parameter. But we need the reference value to to decode relative to! Thus we _must_ have a `RelativeDecoder` trait which has the reference in question as a (`&self`) parameter.
- While it _is_ possible to implement `Encodable` on a Relativity struct (see the first commit), symmetry between `Encodable` and `Decodable` traits has now been broken, which I found a bit confusing
- ~~There is another kind of error exclusive to relative encodings which we'd like to be able to surface, which is `IllegalRelativity`. Some relative encodings require the subject being encoded to be logically included by the reference item, e.g. an `Entry` encoded relative to an `Area` must be included by that `Area`. With a Relativity struct, the only way to check for this would be to define custom constructors for each kind of relative pairing, which was possible to accidentally forget to do. But with a dedicated `RelativeEncoder` trait, we can return `Result<(), RelativeEncodeError>`.~~

All this considered it made more sense to me to have dedicated traits.

## Fuzzing utilities

- `relative_encoding_roundtrip` (encode, decode, check it came out okay)
- `relative_encoding_random` (decode random bytes, encode, assert new encoding is exactly the same as the previous one)
- `relative_encoding_random_less_strict`(decode random bytes, encode, decode it again, assert the decoded object is the same the second time around)

## Encodings

- [x] Path relative to path
- [x] Entry relative entry
- [x] Entry relative area
- [x] Entry relative range
- [x] Area relative area
- [x] 3dRange relative 3dRange


AljoschaMeyer:
> Decoder::decode is a static method which takes a Ufotofu producer as its single parameter. But we need the reference value to to decode relative to! Thus we must have a RelativeDecoder trait which has the reference in question as a (&self) parameter.

Good point. Yeah, I like the two separate traits, then.

> There is another kind of error exclusive to relative encodings which we'd like to be able to surface, which is IllegalRelativity.

Not sure about this one. I think this should be a panic rather than an error, since it indicates a bug in your own implementation rather than an expected case that is allowed to happen.

sgwilym:
> Not sure about this one. I think this should be a panic rather than an error, since it indicates a bug in your own implementation rather than an expected case that is allowed to happen.

Agreed. `RelativeEncodeError` has been removed.

data-model/src/encoding/compact_width.rs
```diff
@@ -88,16 +86,33 @@ impl CompactWidth {
...
+            CompactWidth::One => 0b0000_0000,
+            CompactWidth::Two => 0b0100_0000,
+            CompactWidth::Four => 0b1000_0000,
+            CompactWidth::Eight => 0b1100_0000,
+        };
+
+        og >> position
+    }
+
+    pub fn from_2bit_int(n: u8, position: u8) -> Self {
```
AljoschaMeyer, line 102: Needs a doc comment. I'd *guess* this is the inverse of `bitmask`?

data-model/src/encoding/compact_width.rs
```diff
@@ -88,16 +86,33 @@ impl CompactWidth {
...
+        let mask = 0b0000_0011;
+        let two_bit_int = n >> (6 - position) & mask;
+
+        // Because we sanitise the input down to a 2-bit integer, we can safely unwrap this.
+        CompactWidth::new(2u8.pow(two_bit_int as u32)).unwrap()
+    }
 }
 
 /// Encode a `u64` integer as a `compact_width(value)`-byte big-endian integer, and consume that with a [`BulkConsumer`].
```
AljoschaMeyer, line 111: This *could* be replaced by a wrapper type around u64 that implements the Encoder trait. Not sure whether that makes encoder code significantly nicer.

AljoschaMeyer: Unrelated, how about `Encodable` and `Decodable` instead of `Encoder` and `Decoder` respectively as trait names?

data-model/src/encoding/relativity.rs
```diff
@@ -0,0 +1,281 @@
...
+        &self,
+        reference: &R,
+        consumer: &mut Consumer,
+    ) -> impl Future<Output = Result<(), EncodingConsumerError<Consumer::Error>>>
+    where
+        Consumer: BulkConsumer<Item = u8>;
+}
+
+/// A type that can be used to decode `T` from a bytestring *encoded relative to `Self`*.
+pub trait RelativeDecoder<T> {
```
AljoschaMeyer, line 29: I find it surprising that the RelativeEncoder encodes a value of type `Self`, but the RelativeDecoder does *not* deocde a value of type `Self`. Would switch that for the decoder.

data-model/src/encoding/relativity.rs
```diff
@@ -0,0 +1,281 @@
...
+        reference: &PathRc<MCL, MCC, MPL>,
+        consumer: &mut Consumer,
+    ) -> Result<(), EncodingConsumerError<Consumer::Error>>
+    where
+        Consumer: BulkConsumer<Item = u8>,
+    {
+        let lcp = self.longest_common_prefix(reference);
+        encode_max_power(lcp.component_count(), MCC, consumer).await?;
+
+        let suffix = self.create_suffix(self.component_count() - lcp.component_count());
```
AljoschaMeyer, line 59: Don't love explicit suffix creation, since that involves memory allocation. Would prefer to add a method to Path that iterates over the components but skips over the first `i` (method argument) many.

I can add such a method to the zerocopy paths, can you rework the encodings to work with it when you merge the zerocopy path rewrite?

data-model/src/encoding/relativity.rs
```diff
@@ -0,0 +1,281 @@
...
+        let mut new = prefix;
+
+        for component in suffix.components() {
+            match new.append(component.clone()) {
+                Ok(appended) => new = appended,
+                Err(_) => return Err(DecodeError::InvalidInput),
+            }
+        }
+
+        let actual_lcp = self.longest_common_prefix(&new);
```
AljoschaMeyer, line 98: This way of checking for the error case is elegant, but involves creation of another path value. Which will be pretty cheap with the zerocopy path struct, but still possibly avoidable?

data-model/src/encoding/relativity.rs
```diff
@@ -0,0 +1,281 @@
...
+    }
+}
+
+impl<const MCL: usize, const MCC: usize, const MPL: usize> RelativeDecoder<PathRc<MCL, MCC, MPL>>
+    for PathRc<MCL, MCC, MPL>
+{
+    /// Decode a path relative to this path.
+    ///
+    /// [Definition](https://willowprotocol.org/specs/encodings/index.html#enc_path_relative)
+    async fn relative_decode<Producer>(
```
AljoschaMeyer, line 72: This function will need an efficiency-minded rework once we have the zerocopy path struct.

data-model/src/encoding/relativity.rs
```diff
@@ -0,0 +1,281 @@
...
+
+        if self.timestamp > reference.timestamp {
+            header |= 0b0010_0000;
+        }
+
+        header |= CompactWidth::from_u64(time_diff).bitmask(4);
+
+        header |= CompactWidth::from_u64(self.payload_length).bitmask(6);
+
+        if let Err(err) = consumer.consume(header).await {
```
AljoschaMeyer, line 149: I expect you'll write this pattern of code often enough to warrant turning it into a (family of) functions. So variants of `consume`, `bulk_consume`, `consume_full_slice`, etc that take an additional `consumed_so_far: usize` and that return a result with an `EncodingConsumerError` as the error case. Then you can use those functions with the `?` operator instead of inlining that `if let` construction every time.

data-model/src/encoding/relativity.rs
```diff
@@ -0,0 +1,281 @@
...
+    RelativeDecoder<Entry<N, S, PathRc<MCL, MCC, MPL>, PD>>
+    for Entry<N, S, PathRc<MCL, MCC, MPL>, PD>
+where
+    N: NamespaceId + Decoder + std::fmt::Debug,
+    S: SubspaceId + Decoder + std::fmt::Debug,
+    PD: PayloadDigest + Decoder,
+{
+    /// Decode an [`Entry`] relative from this [`Entry`].
+    async fn relative_decode<Producer>(
+        &self,
```
AljoschaMeyer, line 186: Calling this as `relative_anchor.decode(producer)` feels a bit weird, honestly, since you are not asking the `relative_anchor` to perform any work. Perhaps the trait should define this not as a method but as a function?

data-model/src/encoding/relativity.rs
```diff
@@ -0,0 +1,281 @@
...
+        Producer: BulkProducer<Item = u8>,
+        Self: Sized,
+    {
+        let mut header_slice = [0u8];
+
+        producer
+            .bulk_overwrite_full_slice(&mut header_slice)
+            .await?;
+
+        let header = header_slice[0];
```
AljoschaMeyer, line 199: The previous three lines of code feel again like something worth abstracting into a (small family of) functions.

data-model/src/encoding/relativity.rs
```diff
@@ -0,0 +1,281 @@
...
+
+impl<const MCL: usize, const MCC: usize, const MPL: usize, N, S, PD>
+    RelativeDecoder<Entry<N, S, PathRc<MCL, MCC, MPL>, PD>>
+    for Entry<N, S, PathRc<MCL, MCC, MPL>, PD>
+where
+    N: NamespaceId + Decoder + std::fmt::Debug,
+    S: SubspaceId + Decoder + std::fmt::Debug,
+    PD: PayloadDigest + Decoder,
+{
+    /// Decode an [`Entry`] relative from this [`Entry`].
```
AljoschaMeyer, line 184: Needs link to spec. scrolled here to validate something about the implementation, and was disappointed =(

data-model/src/encoding/relativity.rs
```diff
@@ -0,0 +1,281 @@
...
+    {
+        let mut header_slice = [0u8];
+
+        producer
+            .bulk_overwrite_full_slice(&mut header_slice)
+            .await?;
+
+        let header = header_slice[0];
+
+        // Verify that bit 3 is 0 as specified - good indicator of invalid data.
```
AljoschaMeyer, line 201:
```suggestion
        // Verify that bit 3 is 0 as specified.
```

That second half made me open up the spec and wonder whether I missed anything.

data-model/src/encoding/relativity.rs
```diff
@@ -0,0 +1,281 @@
...
+        Ok(Entry {
+            namespace_id,
+            subspace_id,
+            path,
+            timestamp,
+            payload_length,
+            payload_digest,
+        })
+    }
+}
```
AljoschaMeyer, line 281: I can smell the fuzz testing on this function =D

data-model/src/path.rs
```diff
@@ -151,9 +151,12 @@ pub trait Path: PartialEq + Eq + PartialOrd + Ord + Clone {
         self.component_count() == 0
     }
 
-    /// Create a new [`Path`] by taking the first `length` components of this path.
+    /// Create a new [`Path`] by taking the *first* `length` components of this path.
     fn create_prefix(&self, length: usize) -> Self;
 
+    /// Create a new [`Path`] by taking the *last* `length` components of this path.
+    fn create_suffix(&self, length: usize) -> Self;
```
AljoschaMeyer, line 158: This will not make it into the zerocopy struct, but I will add an iterator over the components of a suffix.

earthstar/src/cinn25519.rs
```diff
@@ -48,9 +48,12 @@ impl<'a, const MIN_LENGTH: usize, const MAX_LENGTH: usize> Arbitrary<'a>
     for Shortname<MIN_LENGTH, MAX_LENGTH>
 {
     fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
-        let shortname_string: String = Arbitrary::arbitrary(u)?;
+        let shortname_bytes: [u8; MIN_LENGTH] = Arbitrary::arbitrary(u)?;
```
AljoschaMeyer, line 51: Some problems here. First, this always generates maximally short (minimally long?) shortnames, now. Generate a `[u8; MAX_LENGTH]` instead, and then a `usize` to indicate how many of those bytes to throw away.

data-model/src/encoding/relativity.rs
sgwilym: So then relative decoder takes two args instead, the thing being decoded relative to, and the producer?

AljoschaMeyer: Yes, see also https://github.com/earthstar-project/willow-rs/pull/24#discussion_r1678237009

But the comment on line 29 was more about the inconsistency between RelativeEncoder and RelativeDecoder.

data-model/src/encoding/compact_width.rs
sgwilym: Better fn names: "unmasking"? `interpret-fixed-width-with-bitmask`? Bitmask should appear in that name. `decode-width-bitmask`?

data-model/src/encoding/relativity.rs
sgwilym: `consume_full_slice` can be automatically turned into an `EncodingConsumerError`.

What is the actual threshold for `consumed_so_far`. Since which point? I'd assume that point is the beginning of an atomic consumption operation began?

AljoschaMeyer: It should be relative to the start of the function call. If I call a function and it returns an EncodingConsumerError, the error should tell me how many bytes that function had consumed, so that I can react accordingly. Most of those may have been part of successful intermediate steps, but as a caller of the function, I should not need to know about them. All I care is that 23 bytes were consumed before the function realised it couldn't decode successfully.

earthstar/src/cinn25519.rs
sgwilym: > I actually prefer the old version (generating a candidate String and running it through Shortname::new). Was there anything wrong with it? The fuzzer would easily manage to generate correct shortnames from that.

```
WARNING: no interesting inputs were found so far. Is the code instrumented for coverage?
This may also happen if the target rejected all inputs we tried so far
```

data-model/src/encoding/relativity.rs
sgwilym: So really I need to only use this new family of byte-counting functions and not use the built-in methods :/ pity

earthstar/src/cinn25519.rs
AljoschaMeyer: Does it start generating output after that message? It sometimes spits out that message but then starts being "productive" after a second or so. I always assumed that was slightly buggy or timer-related behaviour and ignored it.

data-model/src/encoding/relativity.rs
sgwilym: We've decided to simplify the error reporting here.

data-model/src/encoding/error.rs
```diff
@@ -5,19 +5,17 @@ use ufotofu::common::errors::{ConsumeFullSliceError, OverwriteFullSliceError};
 
 /// Returned when a encoding fails to be consumed by a [`ufotofu::local_nb::Consumer`].
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct EncodingConsumerError<E> {
-    /// The number of bytes which were consumed before the error.
-    pub bytes_consumed: usize,
-    /// The error returned on the final and failed attempt to consume bytes.
-    pub reason: E,
-}
+pub struct EncodingConsumerError<E>(E);
```
AljoschaMeyer, line 8: Sorry, I was a bit unclear in the call I guess: you can simply delete this struct and return `Consumer::Error` directly in the encoding functions.

PR #31, opened 2024-07-24
Private entry fields

AljoschaMeyer:
I did not add methods to mutate fields yet. We can add those *if* the need arises.

Not only privatises `Entry`, but also `Range3d` and `Area`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,3 +3,5 @@
 members = ["data-model", "earthstar", "fuzz"]
 resolver = "2"
 
+[workspace.lints.clippy]
+type_complexity = "allow"
--- a/data-model/src/encoding/relativity.rs
+++ b/data-model/src/encoding/relativity.rs
@@ -206,45 +206,45 @@ where
     where
         Consumer: BulkConsumer<Item = u8>,
     {
-        let time_diff = self.get_timestamp().abs_diff(reference.get_timestamp());
+        let time_diff = self.timestamp().abs_diff(reference.timestamp());
 
         let mut header: u8 = 0b0000_0000;
 
-        if self.get_namespace_id() != reference.get_namespace_id() {
+        if self.namespace_id() != reference.namespace_id() {
             header |= 0b1000_0000;
         }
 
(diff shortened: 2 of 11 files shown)


Cargo.toml
```diff
@@ -3,3 +3,5 @@
 members = ["data-model", "earthstar", "fuzz"]
 resolver = "2"
 
+[workspace.lints.clippy]
+type_complexity = "allow"
```
sgwilym, line 7: _thank you_

data-model/src/encoding/relativity.rs
```diff
@@ -283,51 +284,51 @@ where
         let namespace_id = if is_namespace_encoded {
             N::decode(producer).await?
         } else {
-            reference.namespace_id.clone()
+            reference.get_namespace_id().clone()
```
sgwilym, line 287: is `get_whatever` really the norm for this kind of thing? It's so unwieldy compared to `namespace_id` :(

sgwilym:
What does making these fields private get us? To make it impossible to mutate fields after construction?

AljoschaMeyer:
That is not the primary reason. It allows us to later change the definition of the types. We might have an enum where one variant looks like the current struct, but another variant is simply a pointer into a refcounted buffer that contains an encoded version we got from somewhere. Or for `Entry`, we might decide to group path, subspace_id and timestamp together in a substruct that can be used for addressing things in a store.

By using public fields, all those options would require a breaking change.

data-model/src/encoding/relativity.rs
AljoschaMeyer: I had `namespace_id_as_ref` before, and then came to my senses =P

AljoschaMeyer: I can also live with `fn namespace_id` though.

sgwilym: That would resolve my only resistance to this change :)

AljoschaMeyer: Ok, on it.

AljoschaMeyer: Done

PR #32, opened 2024-07-29
Meadowcap

sgwilym:
Implementation of [Meadowcap](https://willowprotocol.org/specs/meadowcap/index.html#communal_cap_receiver).

- Uses traits from [`signature`](https://docs.rs/signature/latest/signature/) crate. 
- Aims to makes it impossible for invalid capabilities to be constructed (i.e. no validation functions)

What's done:

- [x] CommunalCapability
- [x] OwnedCapability
- [x] McCapability
- [x] MeadowcapAuthorisationToken
- [x] McSubspaceCapability
- [x] Encodings
- [x] Tests (always nice)
- [ ] ~~[Extra fancy data structures](https://github.com/earthstar-project/willow-rs/pull/32#discussion_r1695994646)~~ (we will handle this later)


data-model/src/grouping/range.rs
```diff
@@ -118,6 +118,18 @@ impl<const MCL: usize, const MCC: usize, const MPL: usize> PartialOrd<RangeEnd<P
     }
 }
 
+impl<T> Clone for RangeEnd<T>
```
AljoschaMeyer, line 121: Ca this not be derived?

AljoschaMeyer: (Btw, this is another instance of defining a type that is isomorphic to `Option` and then having to implement a bunch of stuff that `Option` already provides.)

data-model/src/grouping/range.rs
```diff
@@ -235,6 +247,18 @@ impl<T: Ord> PartialOrd for Range<T> {
     }
 }
 
+impl<T> Clone for Range<T>
```
AljoschaMeyer, line 250: Can this not be derived?

meadowcap/src/lib.rs
```diff
@@ -0,0 +1,260 @@
+use ufotofu::{local_nb::Consumer, sync::consumer::IntoVec};
+use willow_data_model::{
+    encoding::{parameters::Encodable, relativity::RelativeEncodable},
+    grouping::area::Area,
+    parameters::{NamespaceId, SubspaceId},
+};
+
+/// Can be used to sign a bytestring.
```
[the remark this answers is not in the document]
AljoschaMeyer: Does https://docs.rs/signature/latest/signature/#traits contain anything we could reuse?

meadowcap/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+pub trait Signing<PublicKey, Signature> {
+    fn corresponding_public_key(&self) -> PublicKey;
```
AljoschaMeyer, line 10: While the spec defines things this way, you probably don't want to implement things by calling this function. When an operation needs both a secret key and the corresponding public key, it probably should simply take both as separate arguments. Computing the public key for a secret key is much ore expensive than simply passing around (reference to) both.

meadowcap/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+    fn sign(&self, bytestring: &[u8]) -> Signature;
```
AljoschaMeyer, line 11: https://docs.rs/signature/latest/signature/trait.Signer.html ?

meadowcap/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+}
+
+/// Indicates that this is a verifiable signature.
+pub trait Verifiable<PublicKey> {
+    fn verify(&self, public_key: &PublicKey, bytestring: &[u8]) -> bool;
```
AljoschaMeyer, line 16: https://docs.rs/signature/latest/signature/trait.Verifier.html ?

meadowcap/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+}
+
+/// A delegation of access rights to a user for a given area.
+pub type Delegation<
+    const MCL: usize,
+    const MCC: usize,
+    const MPL: usize,
+    UserPublicKey,
+    UserSignature,
+> = (
```
AljoschaMeyer, line 26: Wouldn't mind a proper struct here. The presentation in the spec has different needs than an implementation.

meadowcap/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+
+/// Returned when an attempt to delegate a capability failed.
+pub enum FailedDelegationError<
+    const MCL: usize,
+    const MCC: usize,
+    const MPL: usize,
+    UserPublicKey: SubspaceId,
+> {
+    /// The granted area of the capability we tried to delegate from did not include the given area.
+    AreaNotIncluded((Area<MCL, MCC, MPL, UserPublicKey>, UserPublicKey)),
```
AljoschaMeyer, line 47:
```suggestion
    AreaNotIncluded(Area<MCL, MCC, MPL, UserPublicKey>, UserPublicKey),
```

AljoschaMeyer, line 47: Or even better:

```suggestion
    AreaNotIncluded {
        helpful_name: Area<MCL, MCC, MPL, UserPublicKey>,
        other_helpful_name_ UserPublicKey,
    },
```

meadowcap/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+/// Returned when an existing delegation was found to be invalid.
+pub enum InvalidDelegationError<
+    const MCL: usize,
+    const MCC: usize,
+    const MPL: usize,
+    UserPublicKey: SubspaceId,
+    UserSignature,
+> {
+    /// The delegation's area was not included by the granted area of the capability we tried to add it to.
+    AreaNotIncluded((Area<MCL, MCC, MPL, UserPublicKey>, UserPublicKey)),
```
AljoschaMeyer, line 61: See `FailedDelegationError::AreaNotIncluded`.

meadowcap/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+    /// The signature of the delegation was not valid for the recevier of the capability we tried to add the delegation to.
+    InvalidSignature((UserPublicKey, UserSignature)),
```
AljoschaMeyer, line 63: See `FailedDelegationError::AreaNotIncluded`.

meadowcap/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+    UserSignature,
+> where
+    NamespacePublicKey: NamespaceId + Encodable,
+    UserPublicKey: SubspaceId + Encodable,
+    UserSignature: Encodable + Verifiable<UserPublicKey>,
+{
+    access_mode: AccessMode,
+    namespace_key: NamespacePublicKey,
+    user_key: UserPublicKey,
+    delegations: Vec<Delegation<MCL, MCC, MPL, UserPublicKey, UserSignature>>,
```
AljoschaMeyer, line 84: This will work for now, but long-term (or even medium-term), we want something that is much cheaper to clone (i.e., something that involves reference counting). We will have many entries with only a few different capabilities, so they should (at least be able to) reuse the same memory. Might even involve [interning](https://en.wikipedia.org/wiki/String_interning). CC @frando, this could be fairly involved. One option would be to require the parameters of the types in a `Delegation` to be [FromBytes](https://docs.rs/zerocopy/0.7.35/zerocopy/trait.FromBytes.html), so that we can store things in a `Bytes` struct.

meadowcap/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+                new_user, new_sig,
+            )));
+        }
+
+        self.delegations.push((new_area, new_user, new_sig));
+
+        Ok(())
+    }
+
+    /// The kind of access this capability grants.
```
AljoschaMeyer, line 182:
```suggestion
    /// The kind of access this capability grants.
    ///
    /// [Definition](https://willowprotocol.org/specs/meadowcap/index.html#communal_cap_mode)
```

meadowcap/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+
+    /// A bytestring to be signed for a new [`Delegation`].
+    ///
+    /// [Definition](https://willowprotocol.org/specs/meadowcap/index.html#communal_handover)
+    // TODO: Make this sync.
+    async fn handover(
+        &self,
+        new_area: &Area<MCL, MCC, MPL, UserPublicKey>,
+        new_user: &UserPublicKey,
+    ) -> Vec<u8> {
```
AljoschaMeyer, line 230: This should probably be a `Box<[u8]>`.

Except that it probably should be neither: how about a bring-your-own-consumer interface where `handover` accepts a (sync) consumer and writes the bytes into it. The current implementation could supply an IntoVec consumer, but later optimised implementations that use reference counted data structures rather than cloning Vecs might call this with a different consumer.

data-model/src/grouping/range.rs
sgwilym: > (Btw, this is another instance of defining a type that is isomorphic to Option and then having to implement a bunch of stuff that Option already provides.)

And here the meaning is isomorphic too. I will refactor to `Option`!

data-model/src/grouping/range.rs
sgwilym: Not without making it so that a Range only accepts values which implement `Clone`.

meadowcap/src/lib.rs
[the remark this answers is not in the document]
sgwilym: Oh, that makes sense, yes!

meadowcap/src/lib.rs
sgwilym: This will change when I adopt the signature crate, but basically I wanted a quick check for determining if a provided secret key was the correct one for a given delegation.

meadowcap/src/lib.rs
sgwilym: At first I had both of these error enums merged as a single one, but that created a situation where users would have to handle an error that would never occur in both cases (e.g. a bad signature verification when creating a new delegation). If that's what you're getting at.

meadowcap/src/lib.rs
sgwilym: See https://github.com/earthstar-project/willow-rs/pull/32#discussion_r1696662583

meadowcap/src/lib.rs
sgwilym: I've always wanted an intern.

But more seriously... let's follow this up in the willow channel on discord.

meadowcap/src/lib.rs
AljoschaMeyer: Nah, I just wanted to say "apply same refactoring as for "FailedDelegationError::AreaNotIncluded" here (removing outer parens or even using named fields). Same for the other variant.

data-model/src/grouping/range.rs
sgwilym: Retracting that refactor after remembering that RangeEnd has a bunch of meaningful semantics on it which prevent its refactoring to an `Option`.

meadowcap/src/lib.rs
sgwilym: Re: bring-your-own-consumer. This is a private function, the only ones bringing anything is `Self` who needs a bytestring for signing or verifying.

sgwilym: Changed the return type to `Box<[u8]>` for now.

meadowcap/src/lib.rs
Frando: Do you mean wrapping the delegations, or the full capability?
In `iroh-willow` I switched to internally wrapping the full delegation in an `Arc`. My reasoning was that cloning capabilities is *much* more frequent than modifying them. With the capability-as-arc, I can just clone them freely whenever and wherever I need them, which is nice.

AljoschaMeyer: I don't have any specific, thought-through ideas. But the same style as the Path struct (encoded data inside a Bytes) comes to mind.
Agreed that cloning is much more important than mutation. Reading a bunch of entries with the same authentication token from the WGPS can reuse refcounted memory when the network encoding specifies the authentication token as one that had been transmitted before. But it might be nice to do interning to also reuse tokens when the network encoding does not explicitly state that it's reusable. Additionally, when reading entries from persistent storage, it would be good to deduplicate memory as well.

sgwilym:
@AljoschaMeyer the fuzz tests are currently running with complaint, but:

- In the case of the appending delegation fuzz test, one path remains unexplored (append success) because the fuzzer can't generate valid (silly) signatures, even though they are, well, silly.
- Generating arbitrary capabilities with many delegations is currently implemented within the fuzz tests rather than within the arbitrary implementations themselves.

I would appreciate if you could inspect the fuzz tests as they are for oversights on my part!

fuzz/fuzz_targets/mc_capability_append_delegation.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    let mut last_receiver = mut_cap.receiver().clone();
+    let granted_area = mut_cap.granted_area();
+
+    for delegee in delegees {
+        mut_cap = mut_cap
+            .delegate(
+                &last_receiver.corresponding_secret_key(),
+                &delegee,
+                &granted_area,
```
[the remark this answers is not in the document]
AljoschaMeyer: Or am I misunderstanding the code here? Hmm...

AljoschaMeyer:
To make the silly signature scheme even more silly, how about defining the signature as a single byte which is the XOR of the public key and the first message byte? That way, there are only 255 different signatures, and the fuzzer should hopefully guess them quickly enough.

fuzz/fuzz_targets/mc_capability_append_delegation.rs
```diff
@@ -0,0 +1,92 @@
...
+                let expected_cap =
+                    mc_cap.delegate(&actual_receiver_secret, &delegation_user, &claimed_area);
+
+                match expected_cap {
+                    Ok(cap) => match cap.delegations().last() {
+                        Some(valid_delegation) => {
+                            assert!(valid_delegation.signature() != &delegation_sig)
+                        }
+                        None => {
+                            unreachable!(
```
AljoschaMeyer, line 82: use `.last.unwrap()` instead of a match expression then. Or `.last().expect("meaningful error")` if there is something meaningful to say.

fuzz/fuzz_targets/mc_capability_append_delegation.rs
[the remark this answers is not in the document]
sgwilym: This phase of the fuzz test is about creating a valid capability with n delegations, which we then test a random secret, user key, and area against.

meadowcap/src/mc_capability.rs
```diff
@@ -0,0 +1,596 @@
...
+};
+
+use crate::{
+    communal_capability::{CommunalCapability, NamespaceIsNotCommunalError},
+    mc_authorisation_token::McAuthorisationToken,
+    owned_capability::{OwnedCapability, OwnedCapabilityCreationError},
+    AccessMode, Delegation, FailedDelegationError, InvalidDelegationError, IsCommunal,
+};
+
+pub struct NotAWriteCapabilityError<
```
AljoschaMeyer, line 18: This should implement `Error`. Alternatively, is it worth considering not having this type at all and simply returning `Result<Yay, McCapability>`?

meadowcap/src/mc_capability.rs
```diff
@@ -0,0 +1,596 @@
...
+where
+    NamespacePublicKey: NamespaceId + Encodable + Verifier<NamespaceSignature> + IsCommunal,
+    UserPublicKey: SubspaceId + Encodable + Verifier<UserSignature>,
+    NamespaceSignature: Encodable + Clone,
+    UserSignature: Encodable + Clone;
+
+/// A Meadowcap capability.
+///
+/// [Definition](https://willowprotocol.org/specs/meadowcap/index.html#Capability)
+#[derive(Clone, Debug, PartialEq, Eq)]
```
AljoschaMeyer, line 46: derive `Arbitrary`?

meadowcap/src/mc_capability.rs
```diff
@@ -0,0 +1,596 @@
...
+        &mut self,
+        delegation: Delegation<MCL, MCC, MPL, UserPublicKey, UserSignature>,
+    ) -> Result<(), InvalidDelegationError<MCL, MCC, MPL, UserPublicKey, UserSignature>> {
+        match self {
+            McCapability::Communal(cap) => cap.append_existing_delegation(delegation),
+            McCapability::Owned(cap) => cap.append_existing_delegation(delegation),
+        }
+    }
+
+    /// Return a new AuthorisationToken without checking if the resulting signature is correct (e.g. because you are going to immediately do that by constructing an [`willow_data_model::AuthorisedEntry`]).
```
AljoschaMeyer, line 210: Is it allowed to produce instances of the type `AuthorisationToken` with incorrect signatures? If not, this should be an `unsafe` function.

meadowcap/src/mc_capability.rs
```diff
@@ -0,0 +1,596 @@
...
+        UserSignature,
+    >
+where
+    NamespacePublicKey:
+        NamespaceId + Encodable + IsCommunal + Arbitrary<'a> + Verifier<NamespaceSignature>,
+    UserPublicKey: SubspaceId + Encodable + Verifier<UserSignature> + Arbitrary<'a>,
+    NamespaceSignature: Encodable + Clone + Arbitrary<'a>,
+    UserSignature: Encodable + Clone,
+{
+    fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
```
AljoschaMeyer, line 568: This could be derived.

meadowcap/src/owned_capability.rs
```diff
@@ -0,0 +1,398 @@
...
+            return Area::new_full();
+        }
+
+        // We can unwrap here because we know delegations isn't empty.
+        let last_delegation = self.delegations.last().unwrap();
+
+        last_delegation.area().clone()
+    }
+
+    /// Return a slice of all [`Delegation`]s made to this capability.
```
AljoschaMeyer, line 286: Would prefer if this returned an `impl Iterator<&Delegation>` so that we can refactor the internals without the limitation of placing delegations successively in memory.

meadowcap/src/subspace_capability.rs
```diff
@@ -0,0 +1,365 @@
...
+        let prev_receiver = self.receiver();
+
+        prev_receiver.verify(&handover, new_sig)?;
+
+        self.delegations.push(delegation);
+
+        Ok(())
+    }
+
+    /// Return a slice of all [`Delegation`]s made to this capability.
```
AljoschaMeyer, line 161: Would prefer if this returned an `impl Iterator<&(PubKey, Sig)>` so that we can refactor the internals without the limitation of placing delegations successively in memory.

meadowcap/src/subspace_capability.rs
```diff
@@ -0,0 +1,365 @@
...
+            NamespaceId + EncodableSync + Decodable + Verifier<NamespaceSignature> + IsCommunal,
+        NamespaceSignature: EncodableSync + Decodable + Clone,
+        UserPublicKey: SubspaceId + EncodableSync + Decodable + Verifier<UserSignature>,
+        UserSignature: EncodableSync + Decodable + Clone,
+    {
+        async fn decode<P>(producer: &mut P) -> Result<Self, DecodeError<P::Error>>
+        where
+            P: BulkProducer<Item = u8>,
+        {
+            // TODO: Strict encoding relations - this decoder will not throw if the number of delegations claimed is not the same as the number of delegations decoded.
```
AljoschaMeyer, line 292: Meadowcap has to be strict about all of these. Encoding *relations* are only an optimisation for the WGPS, but Meadowcap uses encoding *functions* exclusively.

AljoschaMeyer:
Super cool =)!

meadowcap/src/owned_capability.rs
sgwilym: Do you have any recommendations for how to do this with the McCapability, OwnedCapability and CommunalCapability? McCapability::delegations calls upon the delegations method of both these two, but using the `impl Iterator` return type for both of them means they each return a different opaque type.

meadowcap/src/mc_capability.rs
sgwilym: Can you run the latter idea past me once more? If generating `McCapability::authorisation_token` fails, it should return a `McCapability`? Which?

meadowcap/src/owned_capability.rs
AljoschaMeyer: Does the compiler complain about that?

meadowcap/src/mc_capability.rs
AljoschaMeyer: The one that would have been contained in the `NotAWriteCapabilityError`. Or am I missing something? Isn't `NotAWriteCapabilityError` simply a thin wrapper around a `McCapability`?

meadowcap/src/owned_capability.rs
sgwilym: It does.

meadowcap/src/mc_capability.rs
sgwilym: It's not a great return type, as the contained McCapability is a clone of the capability you just called `authorisation_token` on. :/ Maybe I should just remove the wrapped type, and have an empty `NotAWriteCapabilityError`?

AljoschaMeyer: I see. Yeah, that works. Or simply return `Result<Bla, ()>` and not bother with a proper error type?

sgwilym: I need it in one other place where a different kind of error might be returned, so I figured I may as well have it.

meadowcap/src/owned_capability.rs
sgwilym: I can do it by adding _just one more generic_

PR #39, opened 2024-08-09
Implement Error on our error types

sgwilym:
I was pleasantly surprised by how few of these there seem to be. Would like review of clarity of display messages.

### DIFF
--- a/data-model/src/entry.rs
+++ b/data-model/src/entry.rs
@@ -212,7 +212,19 @@ where
 #[derive(Debug)]
 pub struct UnauthorisedWriteError;
 
-/// An AuthorisedEntry is a pair of an [`Entry`] and [`AuthorisationToken`](https://willowprotocol.org/specs/data-model/index.html#AuthorisationToken) (willowprotocol.org) implementing [`AuthorisationToken`] for which [`is_authorised_write`](https://willowprotocol.org/specs/data-model/index.html#is_authorised_write) returns true.
+impl core::fmt::Display for UnauthorisedWriteError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Tried to authorise the writing of an entry using a token which does not permit it."
+        )
+    }
+}
+
+impl std::error::Error for UnauthorisedWriteError {}
+
+/// An AuthorisedEntry is a pair of an [`Entry`] and [`AuthorisationToken`] for which [`Entry::is_authorised_write`] returns true.
+/// [Definition](https://willowprotocol.org/specs/data-model/index.html#AuthorisedEntry).
 ///
 /// [Definition](https://willowprotocol.org/specs/data-model/index.html#AuthorisedEntry).
 pub struct AuthorisedEntry<const MCL: usize, const MCC: usize, const MPL: usize, N, S, PD, AT>(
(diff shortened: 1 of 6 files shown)


data-model/src/entry.rs
```diff
@@ -212,7 +212,19 @@ where
 #[derive(Debug)]
 pub struct UnauthorisedWriteError;
 
-/// An AuthorisedEntry is a pair of an [`Entry`] and [`AuthorisationToken`](https://willowprotocol.org/specs/data-model/index.html#AuthorisationToken) (willowprotocol.org) implementing [`AuthorisationToken`] for which [`is_authorised_write`](https://willowprotocol.org/specs/data-model/index.html#is_authorised_write) returns true.
+impl core::fmt::Display for UnauthorisedWriteError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Tried to authorise the writing of an entry using a token which does not permit it."
```
AljoschaMeyer, line 219:
```suggestion
            "Tried to authorise the writing of an entry using an AuthorisationToken which does not permit it."
```

meadowcap/src/lib.rs
```diff
@@ -136,6 +136,35 @@ pub enum FailedDelegationError<
...
+            FailedDelegationError::AreaNotIncluded {
+                excluded_area: _,
+                claimed_receiver: _,
+            } => write!(
+                f,
+                "Tried to delegate access to an area not fully included by the granted area of the capability being delegated from."
+            ),
+            FailedDelegationError::WrongSecretForUser(_) => write!(
+                f,
+                "Provided the wrong secret for the capability's receiver."
```
AljoschaMeyer, line 153: This sounds like you need the secret key of Alfie in order to delegate to Alfie.

meadowcap/src/lib.rs
```diff
@@ -158,6 +187,45 @@ pub enum InvalidDelegationError<
...
+    > core::fmt::Display for InvalidDelegationError<MCL, MCC, MPL, UserPublicKey, UserSignature>
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            InvalidDelegationError::AreaNotIncluded {
+                excluded_area: _,
+                claimed_receiver: _,
+            } => write!(
+                f,
+                "Tried to append a delegation which granted access to an area not fully included by the granted area of the capability."
```
AljoschaMeyer, line 205:
```suggestion
                "Tried to append a delegation which grants access to an area not fully included by the granted area of the capability."
```

meadowcap/src/lib.rs
```diff
@@ -158,6 +187,45 @@ pub enum InvalidDelegationError<
...
+            ),
+            InvalidDelegationError::InvalidSignature {
+                expected_signatory: _,
+                claimed_receiver: _,
+                signature: _,
+            } => write!(
+                f,
+                "Tried to append a delegation with an invalid signature for the receiver of the target capability."
```
AljoschaMeyer, line 213: This sounds like the receiver has to issue the signature.

meadowcap/src/lib.rs
sgwilym: This is written from the perspective of the capability prior to being delegated, where the receiver is the one who delegates, which seems natural to me.

meadowcap/src/lib.rs
sgwilym: Same thing as https://github.com/earthstar-project/willow-rs/pull/39#discussion_r1712615478

meadowcap/src/lib.rs
sgwilym: As in, I am talking about the *actual* receiver of the capability, not the future capability if this operation were to succeed.

AljoschaMeyer: Still couldn't hurt to clarify this, I guess.

Issue #40, opened 2024-08-09
Move encoding utilities to dedicated crate


sgwilym:
These utilities are currently in the data model crate, and relied upon by that crate + meadowcap. In the future, WGPS will also need access to the utilities in there.

It will be a breaking change, so need to do this sooner rather than later :/

PR #41, opened 2024-08-09
More doc fixes + update README

sgwilym:
@AljoschaMeyer I will do #40 into another branch.

### DIFF
--- a/data-model/src/encoding/traits_sync.rs
+++ b/data-model/src/encoding/traits_sync.rs
@@ -5,7 +5,7 @@ use ufotofu::sync::{BulkConsumer, BulkProducer};
 ///
 /// [Definition](https://willowprotocol.org/specs/encodings/index.html#encodings_what)
 pub trait Encodable {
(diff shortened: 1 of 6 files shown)


data-model/src/encoding/traits_sync.rs
```diff
@@ -29,7 +29,7 @@ pub trait Decodable {
 /// A type that can be used to encode `T` to a bytestring *encoded relative to `R`*.
 /// This can be used to create more compact encodings from which `T` can be derived by anyone with `R`.
 pub trait RelativeEncodable<R> {
-    /// A function from the set `Self` to the set of bytestrings *encoded relative to `reference`*.
+    /// A function which maps the set `(Self, R)` to the set of bytestrings in such a way that any value of `Self` must map to exactly *one* bytestring.
```
AljoschaMeyer, line 32: This still is not enough. That's just the definition of a *function*, not of an *encoding function*. Either put all requirements here, or none. But putting a single requirement there without indication that there are others is asking for trouble.

PR #42, opened 2024-08-09
Move encoding utilities into willow-encoding crate

### LINKED ISSUE #40: Move encoding utilities to dedicated crate

sgwilym:
Closes #40.

A bit of annoying reorganisation which will make re-using these encoding techniques for the WGPS a bit more pleasant and logical.


sgwilym:
@AljoschaMeyer I thought there was something, but I've forgotten now, and I can't for the life of me summon it up. So, um, no?

PR #48, opened 2024-08-22
Add query methods to Store trait

sgwilym:
Adds query trait methods as detailed in https://github.com/earthstar-project/willow-rs/pull/21#issuecomment-2199568852:

> - query
>     - query entry by pair of path and subspace id
>       - yields Entry, AuthorisationToken, available payload prefix (length)
>       - option to filter out results with incomplete payloads
>     - query entries by area (of interest)
>       - all of the functionality of singleton queries, also
>       - ordering results (arbitrary, PTS (optionally reversed), TSP (optionally reversed), SPT (optionally reversed)) (?)
>     - subscriptions: all queries should also work as long-lived subscriptions
>       - this includes subscription to payload prefix appending
>       - this also includes notifications about overwriting and forgetting things
>     - persistent subscriptions
>       - all subscription notifications come with a u64 *progress id*, client code can stop consuming a subscription at any time and can later (even between program shutdown and restart) resume it at the same point by supplying its progress id
>       - this is a best-effort service, the storage may reject a resumption because it is too outdated
>         - a simplistic implementation can effectively opt out of providing persistent subscriptions by simply considering *all* ids as outdated
>       - forgetting entries/payloads can cause progress ids to become outdated, unless the storage tracks the things it forgot
>         - this is why all *forget* functionality comes with a flag whether it should be completely traceless or whether the storage is allowed to track the act of forgetting

To be merged into #21

### DIFF
--- a/data-model/src/lengthy_entry.rs
+++ b/data-model/src/lengthy_entry.rs
@@ -1,4 +1,4 @@
-use crate::{Entry, NamespaceId, PayloadDigest, SubspaceId};
+use crate::{AuthorisationToken, AuthorisedEntry, Entry, NamespaceId, PayloadDigest, SubspaceId};
 
 /// An [`Entry`] together with information about how much of its payload a given [`Store`] holds.
 ///
@@ -29,6 +29,51 @@ where
 
     /// The entry in question.
     pub fn entry(&self) -> &Entry<MCL, MCC, MPL, N, S, PD> {
+        &self.entry
+    }
+
+    /// The number of consecutive bytes from the start of the entry’s Payload that the peer holds.
+    pub fn available(&self) -> u64 {
+        self.available
+    }
+}
+
+/// An [`AuthorisedEntry`] together with information about how much of its payload a given [`Store`] holds.
+pub struct LengthyAuthorisedEntry<
+    const MCL: usize,
+    const MCC: usize,
+    const MPL: usize,
+    N,
+    S,
+    PD,
+    AT,
+> where
+    N: NamespaceId,
+    S: SubspaceId,
+    PD: PayloadDigest,
+    AT: AuthorisationToken<MCL, MCC, MPL, N, S, PD>,
+{
+    /// The Entry in question.
+    entry: AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>,
+    /// The number of consecutive bytes from the start of the entry’s payload that the peer holds.
+    available: u64,
+}
+
+impl<const MCL: usize, const MCC: usize, const MPL: usize, N, S, PD, AT>
+    LengthyAuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>
+where
+    N: NamespaceId,
+    S: SubspaceId,
+    PD: PayloadDigest,
+    AT: AuthorisationToken<MCL, MCC, MPL, N, S, PD>,
+{
+    /// Create a new lengthy entry from a given [`AuthorisedEntry`] and the number of consecutive bytes from the start of the entry’s payload that are held.
+    pub fn new(entry: AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>, available: u64) -> Self {
+        Self { entry, available }
+    }
+
+    /// The entry in question.
+    pub fn entry(&self) -> &AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT> {
         &self.entry
     }
 
--- a/data-model/src/lengthy_entry.rs
+++ b/data-model/src/lengthy_entry.rs
@@ -36,6 +36,23 @@ where
     pub fn available(&self) -> u64 {
         self.available
     }
+
+    /// Turn this into a regular [`Entry`].
+    pub fn into_entry(self) -> Entry<MCL, MCC, MPL, N, S, PD> {
+        self.entry
+    }
+}
+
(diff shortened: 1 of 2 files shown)


data-model/src/lengthy_entry.rs
```diff
@@ -37,3 +37,48 @@ where
...
+
+    /// The entry in question.
+    pub fn entry(&self) -> &AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT> {
+        &self.entry
+    }
+
+    /// The number of consecutive bytes from the start of the entry’s Payload that the peer holds.
+    pub fn available(&self) -> u64 {
+        self.available
+    }
```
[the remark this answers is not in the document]
AljoschaMeyer: Also an AsRef implementation I suppose.

data-model/src/store.rs
```diff
@@ -102,6 +102,44 @@ pub enum PayloadAppendError {
 /// Returned when no entry was found for some criteria.
 pub struct NoSuchEntryError();
```
AljoschaMeyer, line 103:
```suggestion
pub struct NoSuchEntryError;
```

data-model/src/store.rs
```diff
@@ -102,6 +102,44 @@ pub enum PayloadAppendError {
...
 
+/// Orderings for a
+pub enum QueryOrder {
```
AljoschaMeyer, line 106: Not sure we should force all implementations to efficiently support these (or, on the flipside, lead clients to believe that all implementations efficiently support all of these). Not 100% on the correct set of orders, though.

data-model/src/store.rs
```diff
@@ -102,6 +102,44 @@ pub enum PayloadAppendError {
...
+    /// A new entry was ingested.
+    Ingested(u64, AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>),
+    /// An existing entry received a portion of its corresponding payload.
+    Appended(u64, LengthyAuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>),
+    /// An entry was forgotten.
+    EntryForgotten(u64, AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>),
+    /// A payload was forgotten.
+    PayloadForgotten(u64, PD),
+    /// An entry was overwritten.
+    Overwritten(u64, AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>),
```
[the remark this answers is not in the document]
AljoschaMeyer: My gut feeling is to remove this one completely and have only the `Pruned` variant (augmented with the overwriting entry as suggested in Frando's next comment).

data-model/src/store.rs
```diff
@@ -102,6 +102,44 @@ pub enum PayloadAppendError {
...
+    /// An entry was pruned via prefix pruning.
+    Pruned(u64, AuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>),
+}
+
+/// Returned when the store could not resume a subscription because the provided ID was too outdated or not present.
```
AljoschaMeyer, line 140:
```suggestion
/// Returned when the store chooses to not resume a subscription.
```

I feel like stores should be completely free in which subscriptions to resume and which to refuse. No need for us to specify specific circumstances in which they need to do either.

data-model/src/store.rs
```diff
@@ -112,6 +150,8 @@ where
 {
     type FlushError;
     type BulkIngestionError;
+    type EntryProducer: Producer<Item = LengthyAuthorisedEntry<MCL, MCC, MPL, N, S, PD, AT>>;
```
AljoschaMeyer, line 153: I think every method that returns a producer of entries might want its own specific type. So `imply Producer<Bla>` might be better than doing an associated type for each of those.

data-model/src/store.rs
sgwilym: Associated type, maybe?

AljoschaMeyer: That might turn compatibility between different stores into a nightmare, though.

What if we simply did *not* offer any way of specifying an ordering? Ordering is not part of the Willow data model, after all. Although I suppose "newest stuff first" might be quite popular? =/

sgwilym: I thought having orders for each of the dimensions would be fairly uncontroversial.z

No ordering sounds a bit bonkers to me. It's not just newest / oldest being obviously useful, I've relied on path / subspace ordering many times for many directory style interfaces.

data-model/src/store.rs
```diff
@@ -150,6 +150,25 @@ where
 /// Returned when the store chooses to not resume a subscription.
 pub struct ResumptionFailedError(pub u64);
 
+/// Describes which entries to ignore during a query.
+#[derive(Default)]
+pub struct QueryIgnoreParams {
+    /// Omit entries with locally incomplete corresponding payloads.
+    pub ignore_incomplete_payloads: bool,
+    /// Omit entries with an empty payload.
```
AljoschaMeyer, line 158:
```suggestion
    /// Omit entries whose payload is the empty string.
```

Just because I still tend to read this as "Ignore entries for which we have no payload data yet" if I turn off my brain. Not really sure whether my suggestion helps with that, though =S

sgwilym: I sympathise and accept.

AljoschaMeyer:
Lgtm now

PR #49, opened 2024-08-22
Add RbsrStore trait

sgwilym:
As described by https://github.com/earthstar-project/willow-rs/pull/21#issuecomment-2199568852:

> - internal (for replication)
>   - all of the public-facing functionality may also be used internally
>   - query entries by 3dRange
>     - yields Entry, AuthorisationToken, available payload prefix (length)
>     - option to filter out results with incomplete payloads (?)
>     - optionally order according to the [`ReconciliationAnnounceEntries:will_sort`](https://willowprotocol.org/specs/sync/index.html#ReconciliationAnnounceEntriesWillSort) field
>     - optionally as a subscription (?)
>   - summarise 3dRange (count, fingerprint)
>   - convert area of interest to 3dRange

to be merged into #48


sgwilym:
note to self: this trait should be moved into a new WGPS crate.

data-model/src/store.rs
```diff
@@ -301,9 +301,62 @@ where
...
+
+    /// Convert an [`AreaOfInterest`] to a concrete [`Range3d`] including all the entries the given [`AreaOfInterest`] would.
+    fn area_of_interest_to_range(
+        aoi: AreaOfInterest<MCL, MCC, MPL, S>,
+    ) -> Range3d<MCL, MCC, MPL, S>;
+
+    /// Partition a [`Range3d`] into `K` parts, or return `None` if the given range cannot be split into `K` many parts (for instance because the range only includes a single entry).
+    fn partition_range<const K: usize>(
+        range: Range3d<MCL, MCC, MPL, S>,
+    ) -> Option<[Range3d<MCL, MCC, MPL, S>; K]>;
```
Frando, line 361: In `iroh-willow` I have a very similar method (`split_range`) with the notable difference that it returns this:

```rust
    fn split_range(
        &self,
        namespace: NamespaceId,
        range: &Range3d,
        config: &SplitOpts,
    ) -> Result<impl Iterator<Item = Result<RangeSplit>>>;


pub type RangeSplit = (Range3d, SplitAction);

#[derive(Debug)]
pub enum SplitAction {
    SendFingerprint(Fingerprint),
    SendEntries(u64),
}

```

Apart from not using a const generic for `k` and the missing willow generics the semantic difference is that for each subrange, the return value includes a `SplitAction` which either includes the fingerprint of the subrange or the count of entries of the subrange. 

The reasoning is: When this is called from the reconciler, it will *always* do one of the following for each subrange: Call `count` to create a `ReconciliationAnnounceEntries` message, or call `summarise` to get a fingerprint and send a `ReconciliationSendFingerprint`. The store might be able to optimize this if it is looking into the full range already for the `partition_range` call.

AljoschaMeyer: I like the `SplitAction` approach. And I also like returning an iterator, particularly because the store might decide to split into fewer than `k` objects if there are sufficiently few entries.

data-model/src/store.rs
```diff
@@ -301,9 +301,62 @@ where
...
+    S: SubspaceId,
+    PD: PayloadDigest,
+    AT: AuthorisationToken<MCL, MCC, MPL, N, S, PD>,
+{
+    /// Query which entries are [included](https://willowprotocol.org/specs/grouping-entries/index.html#area_include) by a [`Range3d`], returning a producer of [`LengthyAuthorisedEntry`].
+    ///
+    /// If `will_sort` is `true`, entries will be sorted ascendingly by subspace_id first, path second.
+    /// If `ignore_incomplete_payloads` is `true`, the producer will not produce entries with incomplete corresponding payloads.
+    /// If `ignore_empty_payloads` is `true`, the producer will not produce entries with a `payload_length` of `0`.
+    fn query_range(
```
AljoschaMeyer, line 325: Isn't this (and all other methods) missing a `&self` argument?

data-model/src/store.rs
```diff
@@ -301,9 +301,62 @@ where
...
+        range: Range3d<MCL, MCC, MPL, S>,
+        will_sort: bool,
+        ignore_incomplete_payloads: bool,
```
AljoschaMeyer, line 328: `ignore_empty_payloads` is a special case of `ignore_incomplete_payloads`. What happens if you set the prior to true but the latter to false? Would be unambiguous if there was a single argument of an enum type with three options.

Frando: Actually I read it differently:
* ignore_incomplete_payloads will ignore entries where the payload is not fully available locally
* ignore_empty_payloads will ignore entries where payload_len == 0, so entries that have an empty payload

I do think both options make sense. (Likely there could be other options that make sense too, we will see in the future). Two booleans as parameters are quite human error prone in my opinion. I would move to enums. And combine in a `QueryParams` struct, so that it is easier to add more options later:

```rust
pub struct QueryParams {
    pub ignore_incomplete_payloads: bool,
    pub ignore_empty_payloads: bool
}

// add builder methods so you can do
QueryParams::default().ignore_incomplete_payloads()
```

sgwilym: Frando's reading is the same as my intent.

> What happens if you set the prior to true but the latter to false?

You would not get any entries where the payload is not available locally, and entries with an empty payload (which we do have available locally, because we know what the empty payload is) would be included.

AljoschaMeyer: Ah, I see. Makes sense. I also agree that an options struct is less error prone, that would have been my feedback if I hadn't misunderstood the arguments in the first place.

data-model/src/store.rs
sgwilym: Thank you, I like this approach too!

data-model/src/store.rs
sgwilym: I will solve this from #48, and then update here!

PR #51, opened 2024-09-11
Encoding relations API

### LINKED ISSUE #54: Command for running all of the fuzz tests
Given our reliance on fuzz tests, it would be good to have a command which runs all of them in parallel for some amount of time. Useful locally and on CI.

sgwilym:
Willow's site will soon define _encoding relations_ for many encodings, which are a set of acceptable encodings for a given type. These will be useful for things like WGPS messages where we are decoding many messages and we do not want the implicit overhead of verifying that every message was encoded correctly.

We still want to retain the concept of canonical encodings, as these are important for things like Meadowcap signatures.

So this PR adds new `RelationDecodable` and `RelativeRelationDecodable` (open to suggestions) traits with a `decode_relation` and `relative_decode_relation` methods on them, respectively.

The methods have a default implementation which is an alias for `Decodable` / `RelativeDecodable`'s `decode_canonical` and `relative_decode_canonical`.

I made these traits separate so that we can define types which have _only_ a canonical decoder, or _only_ a relation decoder (WGPS messages), or both!

Fuzz tests have been refactored to be synchronous and inclued additional checks on encoding tests.

There is also a new shell script for running all the fuzz tests sequentially (see the README) (closes #54).


encoding/src/traits.rs
```diff
@@ -3,12 +3,16 @@ use std::future::Future;
 use crate::error::DecodeError;
 use ufotofu::local_nb::{BulkConsumer, BulkProducer};
 
-/// A type that can be encoded to a bytestring, ensuring that any value of `Self` maps to exactly one bytestring.
+/// A type which can be asynchronously encoded to bytes consumed by a [`ufotofu::local_nb::BulkConsumer`]
 ///
 /// [Definition](https://willowprotocol.org/specs/encodings/index.html#encodings_what)
 pub trait Encodable {
-    /// Encode a value to a bytestring in a specific way that is best described over at [willowprotocol.org](https://willowprotocol.org/specs/encodings/index.html#encodings_what).
+    /// Encode the value with an [encoding function](https://macromania--channelclosing.deno.dev/specs/encodings/index.html#encoding_function) for the set of `Self`, ensuring that any value of `Self` maps to exactly one bytestring, such that there exists a _decoding_ function such that:
```
AljoschaMeyer, line 10: Is the link to a temporary preview intentional?

encoding/src/traits.rs
```diff
@@ -18,11 +22,16 @@ pub trait Encodable {
...
-    /// Decade a value to a bytestring in a specific way that is best described over at [willowprotocol.org](https://willowprotocol.org/specs/encodings/index.html#encodings_what).
+    /// Decode a bytestring created with an [encoding function](https://macromania--channelclosing.deno.dev/specs/encodings/index.html#encoding_function) for the set of `Self`, ensuring that any value of `Self` maps to exactly one bytestring, such that there exists a _decoding_ function (provided by this trait!) such that:
+    ///
+    /// - for every value `s` in `Self` and every bytestring `b` that starts with `encode_s(s)`, we have `decode_s(b)=s`, and
+    /// - for every `s` in `Self` and every bytestring `b` that does not start with `encode_s(s)`, we have `decode_s(b) ≠ s`.
+    ///
+    /// Will return an error if the encoding correlates to the **encoding relation** and not the result of the aforementioned encoding function.
     ///
     /// [Definition](https://willowprotocol.org/specs/encodings/index.html#decode_s)
     fn decode<Producer>(
```
AljoschaMeyer, line 37: Might want to call this `decode_canonic` to make sure it doesn't get used by accident?

encoding/src/traits.rs
```diff
@@ -46,15 +74,35 @@ pub trait RelativeEncodable<R> {
...
+/// [Definition](https://willowprotocol.org/specs/encodings/index.html#encodings_what)
 pub trait RelativeDecodable<R> {
-    /// Decode a value (relative to a reference value) to a bytestring in a specific way that is best described over at [willowprotocol.org](https://willowprotocol.org/specs/encodings/index.html#encodings_what).
+    ///
+    /// Decode a bytestring created with an [encoding function](https://macromania--channelclosing.deno.dev/specs/encodings/index.html#encoding_function) for the set of `Self` relative to a value of the set of `R`, ensuring that any pair of `Self` and `R` maps to exactly one bytestring, such that there exists a _decoding_ function (provided by this trait!) such that:
+    ///
+    /// - for every pair of `s` in `Self` and `r` in `R`, and every bytestring `b` that starts with `encode_s(s, r)`, we have `decode_s(b, r)=s`, and
+    /// - for every pair of `s` in `Self` and `r` in `R`, and every bytestring `b` that does not start with `encode_s(s, r)`, we have `decode_s(b, r) ≠ s`.
+    ///
+    /// Will return an error if the encoding correlates to the **encoding relation** and not the result of the aforementioned encoding function.
```
AljoschaMeyer, line 87: See above.

AljoschaMeyer (commented):
Looks pretty good, just some small suggestions.

AljoschaMeyer:
> How do we feel about not encoding with encoding relations?

I feel very good about that.

encoding/src/traits.rs
sgwilym: Oops, nope.

encoding/src/traits.rs
```diff
@@ -18,11 +22,16 @@ pub trait Encodable {
...
 ///
 /// [Definition](https://willowprotocol.org/specs/encodings/index.html#encodings_what)
 pub trait Decodable {
-    /// Decade a value to a bytestring in a specific way that is best described over at [willowprotocol.org](https://willowprotocol.org/specs/encodings/index.html#encodings_what).
+    /// Decode a bytestring created with an [encoding function](https://macromania--channelclosing.deno.dev/specs/encodings/index.html#encoding_function) for the set of `Self`, ensuring that any value of `Self` maps to exactly one bytestring, such that there exists a _decoding_ function (provided by this trait!) such that:
+    ///
+    /// - for every value `s` in `Self` and every bytestring `b` that starts with `encode_s(s)`, we have `decode_s(b)=s`, and
+    /// - for every `s` in `Self` and every bytestring `b` that does not start with `encode_s(s)`, we have `decode_s(b) ≠ s`.
+    ///
+    /// Will return an error if the encoding correlates to the **encoding relation** and not the result of the aforementioned encoding function.
```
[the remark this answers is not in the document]
sgwilym: Merci!

encoding/src/traits.rs
sgwilym: So both have explicit names? Yeah, that makes sense to me.

sgwilym:
As I'm working through these encodings, I think I'm going to split off relation decoding into its own trait.

My motivation for this would be to make it possible to implement types _without_ canonical decoders which check for errors. 

For example, it would be a huge waste of time to implement canonical decoders for all of the WGPS message types, when we will only ever decode relations.

Thoughts @AljoschaMeyer?

AljoschaMeyer:
Sounds correct.

PR #57, opened 2024-10-25
WGPS: Add ReadyTransport and friends

sgwilym:
A small function for intercepting the first few bytes of a WGPS session in order to return the session's maximum payload size, received commitment, and transport set to immediately start returning encoded WGPS messages.

### DIFF
--- a/wgps/src/ready_transport.rs
+++ b/wgps/src/ready_transport.rs
@@ -30,14 +30,14 @@ impl<E: core::fmt::Display> core::fmt::Display for ReadyTransportError<E> {
 
 /** The result of intercepting the first few bytes of a WGPS transport. */
 #[derive(Debug)]
-pub struct ReadyTransport<
+pub(crate) struct ReadyTransport<
     const CHALLENGE_HASH_LENGTH: usize,
     E: core::fmt::Display,
     P: Producer<Item = u8, Final = (), Error = E>,
 > {
     /** The maximum payload size which may be sent without being explicitly requested.*/
     pub maximum_payload_size: usize,
-    /** The challenge hash of a nonce */
+    /** The challenge hash of a nonce. */
     pub received_commitment: [u8; CHALLENGE_HASH_LENGTH],
     /** A 'ready' transport set to immediately produce encoded WGPS messages. */
     pub transport: P,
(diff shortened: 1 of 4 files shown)


wgps/src/ready_transport.rs
```diff
@@ -0,0 +1,173 @@
...
+
+/** The result of intercepting the first few bytes of a WGPS transport. */
+#[derive(Debug)]
+pub struct ReadyTransport<
+    const CHALLENGE_HASH_LENGTH: usize,
+    E: core::fmt::Display,
+    P: Producer<Item = u8, Final = (), Error = E>,
+> {
+    /** The maximum payload size which may be sent without being explicitly requested.*/
+    pub maximum_payload_size: usize,
```
AljoschaMeyer, line 39: Not a fan of public fields, would prefer getter methods. Or, as a compromise, can you make the struct `pub(crate)`?

wgps/src/ready_transport.rs
```diff
@@ -0,0 +1,173 @@
...
+/** Given a producer of bytes which is to immediately produce the bytes corresponding to the WGPS' [maximum payload size](https://willowprotocol.org/specs/sync/index.html#peer_max_payload_size) and [received commitment](https://willowprotocol.org/specs/sync/index.html#received_commitment), returns the computed maximum payload size, received commitment, and a 'ready' transport set to produce encoded WGPS messages.
+*/
+pub async fn ready_transport<
+    const CHALLENGE_HASH_LENGTH: usize,
+    E: core::fmt::Display,
+    P: Producer<Item = u8, Final = (), Error = E>,
+>(
+    mut transport: P,
+) -> Result<ReadyTransport<CHALLENGE_HASH_LENGTH, E, P>, ReadyTransportError<E>> {
+    let maximum_payload_power = match transport.produce().await {
```
AljoschaMeyer, line 55: Would be nicer to do `produce().await?` (by implementing `From<E>` for `ReadyTransportError<E>`) I think, assuming you'll need that pattern later as well.

The delayed `?` and the end of the match statement is rather unconventional, and makes reading the code rather weird.

wgps/src/ready_transport.rs
```diff
@@ -0,0 +1,173 @@
...
+    if maximum_payload_power > 64 {
+        return Err(ReadyTransportError::MaxPayloadInvalid);
+    }
+
+    let maximum_payload_size = 2_usize.pow(maximum_payload_power as u32);
+
+    let mut received_commitment = [0_u8; CHALLENGE_HASH_LENGTH];
+
+    for commitment_byte in received_commitment.iter_mut() {
+        match transport.produce().await {
```
AljoschaMeyer, line 72: `.await?` again

AljoschaMeyer (commented):
Looks good, only some small stylistic nitpicks.

PR #77, opened 2025-01-29
Add get_payload to Store trait

sgwilym:
Seems useful.


AljoschaMeyer:
Might need the ability to report an error? I feel like error handling in general in this trait might warrant another look. But for now looks good to merge.

PR #78, opened 2025-02-04
Handle store trait + in memory implementation

sgwilym:
Adds a Handle Store for the binding and retrieval of arbitrary resources with numeric handles.

Also removes some for-now unnecessary crates (earthstar), and includes some changes to LCMUX.


AljoschaMeyer:
Looks good! Will probably need small adjustments with some of the recent spec changes, but seems fine to merge this for now.

PR #79, opened 2025-02-18
Private encodings

sgwilym:
- [x] Private path encoding
- [x] Private area encoding
- [x] Communal capability
- [x] Owned capability
- [x] Subspace capability

### DIFF
--- a/data-model/src/private_encodings/area.rs
+++ b/data-model/src/private_encodings/area.rs
@@ -1,0 +1,368 @@
+use arbitrary::Arbitrary;
+use compact_u64::{CompactU64, Tag, TagWidth};
+use ufotofu_codec::{
+    Blame, Decodable, DecodeError, Encodable, RelativeDecodable, RelativeEncodable,
+};
+use willow_encoding::is_bitflagged;
+
+use crate::{
+    grouping::{Area, AreaSubspace, Range, RangeEnd},
+    Entry, NamespaceId, Path, PayloadDigest, PrivatePathContext, SubspaceId,
+};
+
+#[derive(Debug)]
+/// Confidential data that relates to determining the AreasOfInterest that peers might be interested in synchronising.
+// TODO: Move this all to WGPS?
+pub struct PrivateInterest<
+    const MCL: usize,
+    const MCC: usize,
+    const MPL: usize,
+    N: NamespaceId,
+    S: SubspaceId,
+> {
+    namespace_id: N,
+    subspace_id: AreaSubspace<S>,
+    path: Path<MCL, MCC, MPL>,
+}
+
+impl<const MCL: usize, const MCC: usize, const MPL: usize, N: NamespaceId, S: SubspaceId>
+    PrivateInterest<MCL, MCC, MPL, N, S>
+{
+    fn namespace_id(&self) -> &N {
+        &self.namespace_id
+    }
+
+    fn subspace_id(&self) -> &AreaSubspace<S> {
+        &self.subspace_id
+    }
+
+    fn path(&self) -> &Path<MCL, MCC, MPL> {
+        &self.path
+    }
+
+    pub fn is_more_specific(&self, other: &Self) -> bool {
+        self.namespace_id == other.namespace_id
+            && other.subspace_id.includes_area_subspace(&self.subspace_id)
+            && self.path.is_prefixed_by(&other.path)
+    }
+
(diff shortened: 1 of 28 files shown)


data-model/src/private_encodings/area.rs
```diff
@@ -0,0 +1,368 @@
...
+> {
+    namespace_id: N,
+    subspace_id: AreaSubspace<S>,
+    path: Path<MCL, MCC, MPL>,
+}
+
+impl<const MCL: usize, const MCC: usize, const MPL: usize, N: NamespaceId, S: SubspaceId>
+    PrivateInterest<MCL, MCC, MPL, N, S>
+{
+    fn namespace_id(&self) -> &N {
```
AljoschaMeyer, line 31: Needs to be `pub` I think? Same below.

data-model/src/private_encodings/area.rs
```diff
@@ -0,0 +1,368 @@
...
+    where
+        C: ufotofu::BulkConsumer<Item = u8>,
+    {
+        if !r.rel.includes_area(self) || !r.private.almost_includes_area(self) {
+            panic!(
+                "Tried to encode an Area relative to a PrivateAreaContext it is not included by."
+            )
+        }
+
+        let (start_from_start, end_from_start) = match (r.rel.times().end, self.times().end) {
```
AljoschaMeyer, line 195: Can we do the optimal encoding here? The spec mandates certain choices in some cases, but in the other cases implementations should prefer the choices that yield the smallest encodings. E.g. choosing `let start_from_start = true` if and only if `val.start - rel.start <= rel.end - val.start`.

data-model/src/private_encodings/area.rs
```diff
@@ -0,0 +1,368 @@
...
+
+        if start_from_start {
+            header |= 0b0010_0000
+        }
+
+        if end_from_start {
+            header |= 0b0001_0000
+        }
+
+        let start_diff = match (r.rel.times().end, self.times().end) {
```
AljoschaMeyer, line 220: If selecting start_from_start and end_from_start optimally, these will be computed earlier. Ideally, you'd save one of these paired match statements (doing this same match both for selecting start_from_start and later for computing the value to add/subtract feels a bit inelegant; though I expect an optimising compiler to not care either way, this is only for readability).

data-model/src/private_encodings/area.rs
sgwilym: You can only ever conditionally optimise the closed / closed case here I think, because the value of start_from_start signals whether end_from_start indicates openness or not.

sgwilym: And for `end_from_start`, there is no optimisation, right? The difference of the private area context end from the area end will always be less than the difference from the area start. Right?

AljoschaMeyer: > And for `end_from_start`, there is no optimisation, right? The difference of the private area context end from the area end will always be less than the difference from the area start. Right?

If the context has start 0 and end 1000, and the val has start 0 and end 1, then the absolute difference to the context's start is 1, but the difference to the end is 999.

AljoschaMeyer: > You can only ever conditionally optimise the closed / closed case here I think, because the value of start_from_start signals whether end_from_start indicates openness or not.

start_from_start only signals that if `rel.rel.times.end == open`. If `rel.rel.times.end != open`, than you can optimise here.

data-model/src/private_encodings/area.rs
```diff
@@ -0,0 +1,429 @@
...
+            panic!(
+                "Tried to encode an Area relative to a PrivateAreaContext it is not almost included by."
+            )
+        }
+
+        let (start_from_start, end_from_start) = match (r.rel.times().end, self.times().end) {
+            (RangeEnd::Open, RangeEnd::Open) => (true, false),
+            (RangeEnd::Open, RangeEnd::Closed(_)) => (true, true),
+            (RangeEnd::Closed(rel_end), RangeEnd::Open) => {
+                let start_from_start =
```
AljoschaMeyer, line 211: Silly question: is this branch even reachable?

AljoschaMeyer: Answer: yes, the spec technically allows it. But the WGPS will never trigger this branch.

PR #81, opened 2025-03-12
Make bulk_ingest_entry a default method

AljoschaMeyer:
This turns your implementation of `bulk_ingest_entry` on the simple sled store into a default method of the `Store` trait.

To do so without introducing some `: Clone` bounds, I simplified the return type. There errors are not paired with the authorised entry which raised them, instead the caller should match errors to entries based on the order in which they are emitted. Does that work for you? The alternative - gating implementation of bulk ingestion behind Clone bounds - felt a bit awkward to me.

### DIFF
--- a/data-model/src/store.rs
+++ b/data-model/src/store.rs
@@ -4,6 +4,7 @@ use std::{
     future::Future,
 };
 
+use either::Either;
 use ufotofu::{BulkConsumer, BulkProducer, Producer};
 
 use crate::{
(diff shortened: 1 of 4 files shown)


AljoschaMeyer:
(You can ignore all code changes in the fuzz crate, from your perspective this PR touches the store trait and the simple sled store only)

sgwilym (approved):
looks great, thank you :)

PR #84, opened 2025-03-12
No more implicit content addressing!

sgwilym:
As discussed. Massively simplifies things in the WIP implementation of `StoreSimpleSled`.

### DIFF
--- a/store_simple_sled/src/lib.rs
+++ b/store_simple_sled/src/lib.rs
@@ -1,24 +1,20 @@
 use either::Either;
-use std::{collections::HashMap, convert::Infallible, hash::Hash, marker::PhantomData};
+use std::convert::Infallible;
 
 use sled::{
     transaction::TransactionalTree, Db, Error as SledError, IVec, Result as SledResult,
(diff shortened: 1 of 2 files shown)


store_simple_sled/src/lib.rs
```diff
@@ -414,73 +325,98 @@ where
...
+                    return Err(PayloadAppendError::TooManyBytes);
+                }
+
+                let existing_payload = payload_tree
+                    .get(&key)
+                    .map_err(|_| PayloadAppendError::OperationError(SimpleStoreSledError {}))?;
 
-        // Ingest
+                let prefix = if let Some(payload) = existing_payload {
+                    if payload.len() as u64 == length {
```
AljoschaMeyer, line 362: Is this explicit check necessary? Appending zero bytes to a complete payload should arguably be allowed, and if the expected length is being exceeded, that's caught later.

I didn't read the full context of this line, so you might want to ignore this comment if there are good reasons for this check.

AljoschaMeyer:
Looks good!

PR #85, opened 2025-03-13
Remove notion of store operations

sgwilym:
Removing the notion of store operations and traceless operations from the core `Store` trait (though may re-introduce them in a supertrait later).


AljoschaMeyer:
Looks ok.

Issue #56, opened 2024-09-25
Tracking: Ideas for useful helper functions


matheus23:
I'll try keeping track of some useful helper functions to include in `willow-rs`.

- [ ] Some form of `Range::new_exact`:
  ```rs
  impl Range<Path> {
    fn new_exact_path(path: Path) -> Self;
  }
  ```
  
  An alternative for this would be:
  ```rs
  trait Successor { // which is actually equivalent to `trait SubspaceId` at this point.
    fn successor(&self) -> Option<Self>;
  }
  
  impl<T: Successor> Range<T> {
    fn new_exact(value: T) -> Self;
  }
  ```
- [ ] Helpers for creating `Path`s from arrays of bytes/strings:
  ```rs
  impl Path {
    fn from_bytes(&[&[u8]]) -> Self;
    fn from_strs(&[&str]) -> Self;
  }
  ```
- [ ] Helpers for printing `Path`s. Perhaps just a nice `Path: Debug` instance that interprets the path segments as utf8, if possible, or otherwise escapes to hex or similar. Alternatively something like `Path::to_string(&self) -> Option<String>` that tries to format using `/` and utf8, but fails if the path would have to be escaped.
- [ ] Easily converting an `Area` into a `Range3d`: `Area::to_range3d(&self) -> Range3d`. Alternatively: a `From`/`Into` instance.
- [ ] Easily converting an `Area` into an `AreaOfInterest`: `AreaOfInterest::new_area(area: Area) -> Self` or similar

sgwilym:
Thank you for these suggestions, I'm agreed on all counts. I like the `Successor` trait, I think the only drawback would be needing a newtype to implement it on `u64` for timestamps.

matheus23:
FWIW, I personally find it *just* easy enough to do a `Range::new_closed`, if it's a `Range<u64>`, so I'd be fine without a newtype.

PR #102, opened 2025-04-09
Add Path::from_slices helper function

nicholastmosher:
Tackling part of https://github.com/earthstar-project/willow-rs/issues/56

I went with `Path::from_slices` rather than separate `Path::from_{bytes,strs}` because I didn't want to duplicate the logic, but there wasn't really a clean way to implement one in terms of another because even though `&str` can dereference to `&[u8]`, it's not possible to convert `&[&str]` to `&[&[u8]]` without allocating or transmuting (as far as I could see).

I saw that other methods have a `#### Complexity` section but I didn't feel confident in determining what the total complexity was because I can't quite tell how `append_component` behaves (complexity-speaking).

Also let me know if or where the best place to add a unit test for this would be!

I'm running late for something so I'll drop this and check back later!

### DIFF
--- a/data-model/src/path/mod.rs
+++ b/data-model/src/path/mod.rs
@@ -536,17 +536,64 @@ impl<const MCL: usize, const MCC: usize, const MPL: usize> Path<MCL, MCC, MPL> {
     }
 
     /// Create a new path from a slice over byte slices.
-    pub fn from_slices<T: AsRef<[u8]>>(slices: &[T]) -> Result<Self, InvalidPathError> {
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use willow_data_model::{Path, PathConstructionError, InvalidPathError};
+    /// // Ok
+    /// let path = Path::<12, 3, 30>::from_slices(&["alfie", "notes"]).unwrap();
+    ///
+    /// // Err
+    /// let result1 = Path::<12, 3, 30>::from_slices(&["themaxpathlengthis30andthisistoolong"]);
+    /// assert!(matches!(result1, Err(PathConstructionError::InvalidPath(InvalidPathError::PathTooLong))));
+    ///
+    /// // Err
+    /// let result2 = Path::<12, 3, 30>::from_slices(&["too", "many", "components", "error"]);
+    /// assert!(matches!(result2, Err(PathConstructionError::InvalidPath(InvalidPathError::TooManyComponents))));
+    ///
+    /// // Err
+    /// let result3 = Path::<12, 3, 30>::from_slices(&["overencumbered"]);
+    /// assert!(matches!(result3, Err(PathConstructionError::ComponentTooLongError)));
+    /// ```
+    pub fn from_slices<T: AsRef<[u8]>>(slices: &[T]) -> Result<Self, PathConstructionError> {
         let total_length = slices.iter().map(|it| it.as_ref().len()).sum();
         let mut builder = PathBuilder::new(total_length, slices.len())?;
 
         for component_slice in slices {
             let component = Component::<MCL>::new(component_slice.as_ref())
-                .ok_or(InvalidPathError::ComponentTooLong)?;
+                .ok_or(PathConstructionError::ComponentTooLongError)?;
             builder.append_component(component);
         }
 
         Ok(builder.build())
+    }
+}
+
+/// Raised upon failing to construct a [`Path`] from a slice of slices.
+#[derive(Debug, Clone)]
+pub enum PathConstructionError {
+    InvalidPath(InvalidPathError),
+    ComponentTooLongError,
+}
+
+impl std::error::Error for PathConstructionError {}
+impl std::fmt::Display for PathConstructionError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            PathConstructionError::InvalidPath(invalid_path_error) => {
+                std::fmt::Display::fmt(invalid_path_error, f)
+            }
+            PathConstructionError::ComponentTooLongError => {
+                write!(f, "failed to construct a `Path` from slices")
+            }
+        }
+    }
+}
+
+impl From<InvalidPathError> for PathConstructionError {
+    fn from(value: InvalidPathError) -> Self {
+        Self::InvalidPath(value)
     }
 }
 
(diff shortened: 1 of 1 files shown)


AljoschaMeyer:
Nice, thank you!

> I went with `Path::from_slices` [...]

Going with `T: AsRef<[u8]>` is a really good choice 👍 

> I can't quite tell how append_component behaves (complexity-speaking)

Good point, I'll document the complexity of the `PathBuilder` methods.

> Also let me know if or where the best place to add a unit test for this would be!

Unit tests can simply go in a test module in the file of the function they are testing.

We have a quite elaborate [fuzz testing](https://rust-fuzz.github.io/book/introduction.html) setup for paths that checks the equivalence of our optimised `Path` struct and a simplistic reference path struct. *If* you want to go all in, the steps are:

- Add a `from_slices` method to [`PathRc`](https://github.com/earthstar-project/willow-rs/blob/main/fuzz/src/path.rs#L165)
- Add [unit tests](https://github.com/earthstar-project/willow-rs/blob/main/fuzz/src/path.rs#L364) for `PathRc::from_slices`
- Add a `FromSlices` variant to [`CreatePath`](https://github.com/earthstar-project/willow-rs/blob/main/fuzz/src/path.rs#L780) (containing a `Vec<Vec<u8>>`, probably)
- Add and implement a corresponding match case in [`create_path_rc`](https://github.com/earthstar-project/willow-rs/blob/main/fuzz/src/path.rs#L790) and in [`create_path`](https://github.com/earthstar-project/willow-rs/blob/main/fuzz/src/path.rs#L853)
- Execute each of `cargo fuzz run path`, `cargo fuzz run path2`, `cargo fuzz run path3`, `cargo fuzz run path4` and check that the fuzzer is unable to crash them within a couple of minutes

But this is probably overkill: we built this fuzz system mostly to ensure that we didn't mess up the byte-level fiddling in the internal optimised path representation. But since we have the `PathBuilder` now, we can be pretty confident that things just work. So in my book, a couple of unit tests in `src/path/mod.rs` are completely sufficient.

data-model/src/path/mod.rs
```diff
@@ -534,6 +542,20 @@ impl<const MCL: usize, const MCC: usize, const MPL: usize> Path<MCL, MCC, MPL> {
 
         None
     }
+
+    /// Create a new path from a slice over byte slices.
```
AljoschaMeyer, line 546:
```suggestion
    /// Create a new path from a slice of byte slices.
    ///
    /// #### Complexity
    ///
    /// Runs in `O(n + m)`, where `n` is the total length of the path in bytes, and `m` is the number of components. Performs a single allocation of `O(n + m)` bytes.
```

AljoschaMeyer: I realise that some of the other complexity sections on `Path` are incorrect, because they ignore the number of components. I will fix those in a separate pull request.

data-model/src/path/mod.rs
```diff
@@ -25,8 +25,10 @@ mod codec; // Nothing to import, the file only provides trait implementations.
 pub use codec::{decode_path_extends_path, encode_path_extends_path};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
-/// An error arising from trying to construct a invalid [`Path`] from valid components.
```
AljoschaMeyer, line 28: I'd prefer to have a new error enum for methods that work with possibly invalid component data, and leave the old `InvalidPathError` unchanged. If the old error type is merely generalised, than callers of functions that cannot have the failure mode of invalid components (say, [`Path::new_singleton`](https://github.com/earthstar-project/willow-rs/blob/main/data-model/src/path/mod.rs#L94)) still need to devote a `InvalidPathError::ComponentTooLong => unreachable!()` branch whenever handling the error. Which somewhat defeats the point of statically typed errors.

nicholastmosher: Considering that `Path::from_slices` has three error states (both variants from `InvalidPathError` from `PathBuilder::new` plus now a "ComponentTooLong" possibility from `Component::new` used on each subslice.

What would you prefer here? One idea I have is something like this:

```rust
pub enum PathConstructionError {
    InvalidPath(InvalidPathError),
    ComponentTooLong,
```

This provides exactly one case for each way the construction can fail, so it seems fitting.

nicholastmosher:
I added a doctest, I like doctests but I know not everybody does, let me know if you'd prefer the test in a `#[cfg(test)]` module instead.

data-model/src/path/mod.rs
AljoschaMeyer: Yes, I think this type is the way to go. Also a solid name 👍 (I failed to come up with one I was happy with before...)

data-model/src/path/mod.rs
```diff
@@ -534,6 +534,71 @@ impl<const MCL: usize, const MCC: usize, const MPL: usize> Path<MCL, MCC, MPL> {
...
+
+impl std::error::Error for PathConstructionError {}
+impl std::fmt::Display for PathConstructionError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            PathConstructionError::InvalidPath(invalid_path_error) => {
+                std::fmt::Display::fmt(invalid_path_error, f)
+            }
+            PathConstructionError::ComponentTooLongError => {
+                write!(f, "failed to construct a `Path` from slices")
```
AljoschaMeyer, line 592: Since this error type might be used by other functions in the future as well, this could use a more generic description.

```suggestion
                write!(f, "Length of a path component in bytes exceeded the maximum component length"")
```

AljoschaMeyer:
> I added a doctest, I like doctests but I know not everybody does, let me know if you'd prefer the test in a `#[cfg(test)]` module instead.

Yay =). I'm a fan of good doc tests (and other parts of the code base already use them, the path code is merely happens to be too old for that — and too well fuzzed)

AljoschaMeyer:
If you are fine with my suggested changes, this would be ready for merging unless I think ^_^

nicholastmosher:
I just applied those suggestions, so if it looks good to you it looks good to me! 😁

AljoschaMeyer:
I've also updated the complexity section of a bunch of methods to reflect the number of components now.

PR #104, opened 2025-04-15
Add working example for using StoreSimpleSled

sgwilym:
To both the README and module level docs (where it passes doctest!)

### DIFF
--- a/store_simple_sled/src/lib.rs
+++ b/store_simple_sled/src/lib.rs
@@ -6,6 +6,24 @@
 //! - *Simple*, hence it has a straightforward implementation without the use of fancy data structures.
 //! - Uses [sled](https://docs.rs/sled/latest/sled/) under the hood.
 //!
+//! ```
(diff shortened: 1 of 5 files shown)


store_simple_sled/src/lib.rs
```diff
@@ -6,6 +6,24 @@
...
+//!
+//! let store = StoreSimpleSled::<
+//!     1024,
+//!     1024,
+//!     1024,
+//!     NamespaceId25,
+//!     SubspaceId25,
+//!     PayloadDigest25,
+//!     AuthorisationToken25
+//! >::new(&namespace, db).unwrap();
```
AljoschaMeyer, line 24: Couldn't hurt to perhaps insert an entry, add its payload, and then retrieve its payload (and assert equality).

sgwilym: With these parameters that involves creating a keypair, a capability, generating an entry, making an auth token for it - all a bit much for this example.

PR #116, opened 2025-06-14
Add Entry::is_pruned_by method

sgwilym:


AljoschaMeyer:
Should also have `prunes` (implemented by calling `is_pruned_by` with reversed arguments).

PR #117, opened 2025-06-16
Add SubspaceId::singleton_range, Path::singleton_range

sgwilym:


AljoschaMeyer:
Suuuuper nitpickey, but in the doc comments, "exactly one" instead of "only one" would be less ambiguous.
I'm not proud that I'm feeling strongly about this, but I do...

data-model/src/parameters.rs
```diff
@@ -18,6 +19,17 @@ pub trait NamespaceId: Eq + Default + Clone + Debug {}
 pub trait SubspaceId: Ord + Default + Clone + Debug {
     /// Returns the next possible value in the set of all [`SubspaceId`], i.e. the (unique) least value that is strictly greater than `self`. Only if there is no greater value at all may this method return `None`.
     fn successor(&self) -> Option<Self>;
+
+    /// Returns a range which *only* [includes](https://willowprotocol.org/specs/grouping-entries/index.html#range_include) the value of `self`.
```
AljoschaMeyer, line 23:
```suggestion
    /// Returns a range which [includes](https://willowprotocol.org/specs/grouping-entries/index.html#range_include) **exactly** one value, which is that of `self`.
```

PR #119, opened 2025-06-19
McCapability: Check namespace and granted area when creating AuthorisationToken

sgwilym:


AljoschaMeyer:
> I'm confused. Why would an entry be checked against a cap's granted namespace and area when a cap is created?

Uhm, yeah, never mind...

PR #126, opened 2025-07-30
Add DecodedMcEnumCap with RelativeDecodable impl

sgwilym:
- Implements RelativeEncodable for UnverifiedMcEnumerationCapability  
- Implements RelativeDecodable for DecodedEnumCap 
- No fuzz tests as it's not possible to use our fuzzing utilities where type A is decoded to type B.


AljoschaMeyer:
Thanks!

PR #130, opened 2025-08-14
Sideloading

sgwilym:
This PR adds a new crate, `willow-sideload`, which implements the [Sideloading protocol](https://macromania--macromania.deno.dev/specs/sideloading/index.html#sideloading)(preview link).

The API in a nutshell:
- `create_drop`, which you give an iterator of `Area`s to ingest entries from, a `Store`, and a ufotofu `Consumer` with which to consume the encoded bytes of the resulting drop.
- `ingest_drop`, which you give an ufotofu producer of bytes (i.e. the code of the drop), and a `Store` which should ingest the encoded entries and payloads.

I've also got a placeholder Willow'25 wrapper for these APIs which configures everything **except** for encryption / decryption, so it is not ready for use yet.


AljoschaMeyer:
Looks good, ready to merge for me!

PR #137, opened 2025-09-05
RbsrStore implementation for StoreSimpleSled + ControlStore

sgwilym:


AljoschaMeyer:
For reasons that are not at all cursed, I had to change some Store and RbsrStore methods to take `self` as an `Rc<Self>` instead  of a `&Self`, and also to take some other arguments by value. This is so that the returned `impl Producer`s are `'static`. @sgwilym, could you update the method impls accordingly on this branch? Changes introduced at https://github.com/earthstar-project/willow-rs/commit/fa94d8b4778dfde6c200693585ec0043027e15d3.

AljoschaMeyer:
PR looks otherwise good to merge to me at a rough glance.

